Issue #2, opened 2017-09-06
Tooling


fczuardi:
One tool that I like a lot for fixing formatting/code-style/lint issues is [prettier](https://github.com/prettier/prettier), maybe we could replace eslint and the so called "standard" js with it.

I would like to experiment with [Jest](https://facebook.github.io/jest/) as well for tests, since we already have it for storyshots.

Other tool that could be part of our pipeline is a type checker, the one I like is [flow](https://flow.org/).

fczuardi:
We could organize our commit messages and our changelog + release versions using [commitzen](https://github.com/commitizen/cz-cli) and [semantic-release](https://github.com/semantic-release/semantic-release). I have never used them in a project before, but the idea of having auto generated changelogs and not having to remember to bump the correct semver numbers sounds good...

fczuardi:
Issue #24 is opened to track lerna publish and semantic release work

fczuardi:
prettier, jest, flow and commitzen are running on the monorepo branch, as well as lint-staged but that one still have an issue with partial commits to be resolved (see @alcferreira comment on the pr #19 )

alcferreira:
I'm closing this for now... We can reopen this later if necessary (:

PR #19, opened 2017-09-27
Monorepo

### LINKED ISSUE #3: Organize this project as a monorepo
Since this is a collection of multiple widgets/components, the [unix philosophy](https://en.wikipedia.org/wiki/Unix_philosophy) would suggest that each component should be published on npm as its own single library, but that can become very impractical very fast, and we dont want that.

The monorepo approach can be a way for us to have single packages without raising maintainance overhead, this bug is for discussing the topic and coordinating actionable steps we can take to implement that approach in here.

- https://yarnpkg.com/blog/2017/08/02/introducing-workspaces/
- https://github.com/babel/babel/blob/master/doc/design/monorepo.md

fczuardi:
closes #3 and maybe #2 

IMPORTANT: before merging this branch with master, merge other individual packages branches into this monorepo branch. For exemple, merge Pr #20 before merging this. The idea is for the monorepo branch to replace / become the master branch eventually.


alcferreira:
Hmmm... I'm not sure about this `precommit` script.
I like it but we may commit something that is not intended.

Let's assume that I've changed a file and for now I just want to commit some of these changes.
So I run `git add --patch` and choose the changes I want to commit.

So far so good.

Then `precommit` will fix my code formating and will add all of my changes to the commit (not just the ones that I've staged)... 

What are your thoughts? 
@fczuardi @henlz

alcferreira:
Also see this [comment](https://github.com/CraveFood/farmblocks/pull/20#issuecomment-332622277)

henlz:
It doesn't look so clear to me about this issue with adding the formatted files on the precommit hook. If it indeed adds the files, then I have to agree with @alcferreira on this matter.

fczuardi:
history cleanup, rebased from 44 to 21 commits

fczuardi:
master has coveralls integration, I am opening #39 to keep track of integrating that here as well

flow-typed/npm/prop-types_v15.x.x.js
```diff
@@ -0,0 +1,35 @@
...
+declare module "prop-types" {
+  declare var array: React$PropType$Primitive<Array<any>>;
+  declare var bool: React$PropType$Primitive<boolean>;
+  declare var func: React$PropType$Primitive<Function>;
+  declare var number: React$PropType$Primitive<number>;
+  declare var object: React$PropType$Primitive<Object>;
+  declare var string: React$PropType$Primitive<string>;
+  declare var any: React$PropType$Primitive<any>;
+  declare var arrayOf: React$PropType$ArrayOf;
+  declare var element: React$PropType$Primitive<any>; /* TODO */
```
henlz: why TODO?

fczuardi:
Those libdefs are downloaded from the internet with the command 'flow-typed install package@version' I didn't wrote them :)

-------- Original Message --------

fczuardi:
comment https://github.com/CraveFood/farmblocks/pull/19#issuecomment-332618737 addressed by PR #44 pre-commit handles partially staged files better than the option 1 from the prettier documentation (lint staged configured for git-adding files after lint fix)

flow-typed/npm/prop-types_v15.x.x.js
fczuardi: see [below](https://github.com/CraveFood/farmblocks/pull/19#issuecomment-334940377)

fczuardi:
👍

PR #36, opened 2017-10-04
Farmblocks button

### LINKED ISSUE #25: merge Button in monorepo branch
this issue is to target the necessary work to bring the existing button component into the monolithic repo

dependes on:

- [x] #38
- [x] #32

### LINKED ISSUE #31: type anotate Button component main file
the current Button.js file dont start with ``// @flow`` and is kept outside of the typechecks

this is a small refactoring issues just to enable typechecking for that file and make the necessary adjusts to make it pass the checks

### LINKED ISSUE #38: temporarily include the whatsgood iconfont on the storybook webpack of the button component
While the Button component don't use the Icon component (issue #35), we should improve the visuals of the button storybook by including the font family css, with the classname rule and the whatsgood iconfont files in a static folder to be served by this development local webpack server.

fczuardi:
- closes #25
- closes #31 
- closes #38


fczuardi:
branch ready for code review

packages/button/src/constants/SizeTypes.js
```diff
@@ -0,0 +1,3 @@
+export const SMALL = "small";
+export const MEDIUM = "medium";
```
alcferreira: Remove this `MEDIUM` type since we don't use it anywhere

packages/button/src/Button.js
```diff
@@ -60,28 +69,23 @@ class Button extends Component {
     );
   }
 
-  mouseOver() {
+  mouseOver = function() {
```
alcferreira: Any reasons to change `mouseOver()` to `mouveOver = function()`?
Personally i prefer `mouseOver()`

packages/button/src/Button.js
```diff
@@ -0,0 +1,110 @@
...
+        </StyledButton>
+        <Tooltip
+          text="This action is disabled."
+          isVisible={this.state.showTooltip}
+        />
+      </Container>
+    );
+  }
+
+  mouseOver = function() {
```
alcferreira: Any reasons to change `mouseOver()` to `mouveOver = function()`?
Personally i prefer `mouseOver()`... On webapp we go with this syntax

alcferreira:
On storybook the `icon only` button has the same dimensions for both `small` and `large` sizes.
Can you check?

image

image

packages/button/src/Button.js
fczuardi: Flow will complain about covariance otherwise, (see https://github.com/facebook/flow/issues/3076 ) but there is an alternative solution that might be more clean for us to use, that is to declare the Function types like this: https://github.com/ryyppy/flow-guide/issues/6#issuecomment-265398623

I will give that approach a try.

PR #42, opened 2017-10-05
Root storybook + coveralls

### LINKED ISSUE #30: find a way to reuse a centralised storybook config on all packages
having the same config on all packages doent make sense, we should centralise storybook config in the root of the monorrepo and use the same one for all packages.

### LINKED ISSUE #39: setup coveralls on the monorepo branch
this should be simple, we need to add --coverage to the jest test tasks and matbe some coveralls config that we can bring from the master branch

fczuardi:
- [x] closes #30 
- [x] closes #39


packages/tooltip/package.json
```diff
@@ -11,34 +13,27 @@
...
     "@storybook/react": "^3.2.12",
     "babel-cli": "^6.26.0",
-    "flow-bin": "^0.55.0",
-    "flow-typed": "^2.1.5",
-    "jest": "^21.1.0"
+    "flow-bin": "^0.56.0",
+    "jest": "^21.2.1"
   },
-  "keywords": []
+  "keywords": ["farmblocks", "tooltip", "react", "comopnent"]
```
henlz: typo on `comopnent`

jest.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  verbose: true,
+  setupFiles: ["<rootDir>/jest.setup.js"],
+  coveragePathIgnorePatterns: ["jest.setup.js", ".story.js"],
+  testPathIgnorePatterns: ["src/Storyshots.test.js"]
```
henlz: are you sure that this `testPathIgnorePatterns` property is working?

It seems that jest is trying to run the `Storyhots.test.js` file

Anyway, why would we ignore this file?

screen shot 2017-10-09 at 15 57 35

fczuardi: this import bug is a yarn / jest / babel thing, and the reason I've added ``yarn add -D jest`` on travis after all install steps. really strange and we might need to report this on the proper projects (probably on jest first)

if you add jest again it starts passing the test, and the most curious is that the lockfile dont change, remains the eexact same dependency tree

Issue #41, opened 2017-10-05
generate full static storybook and publish on the web as documentation / gallery


fczuardi:
Maybe related with #24 

Storybooks have a function that generates a static version of the components gallery (storybook buld), we should keep an updated version of this site, containing all released versions of all comopnents, published on the web.

The easiest/fastest way of doing this would be to commit a ``docs`` folder on the master branch and serve it up with github pages.

This is something that a Travis job could do for us when new patches land on master.

PR #45, opened 2017-10-09
docs(gallery): create a gallery page under docs to be served by githu…

fczuardi:
…b pages

in the future we would want travis to do this for us, but for now, I am commiting this folder
manually, it is the result of the build-storybook command

ISSUES CLOSED: #41

PR #51, opened 2017-10-16
docs(gallery): create a gallery page under docs to be served by githu…

fczuardi:
…b pages

(old PR #45)

in the future we would want travis to do this for us, but for now, I am commiting this folder
manually, it is the result of the build-storybook command

ISSUES CLOSED: #41


henlz:
What about the broken icons?
screen shot 2017-10-17 at 15 11 25

fczuardi:
fixed the assets

PR #57, opened 2017-10-18
[WIP] Auto generate React propTypes from flow annotations

### LINKED ISSUE #22: Auto generate React propTypes from flow annotations
By using a babel plugin ([babel-plugin-flow-react-proptypes](https://github.com/brigand/babel-plugin-flow-react-proptypes)) we might be able to have both the benefits of flow typechecking and React runtime type checking without the need of duplicate definitions on our code (or annotating props arguments with a generic Object type).

The other alternative would be to use [reason-react](https://github.com/reasonml/reason-react), but that might be a bigger jump for the short term, and should be filed later as a different issue.

alcferreira:
Closes #22


fczuardi:
After a quick call with @alcferreira we opt to keep this effort on hold for now and follow with the manual proptypes writing for a little longer since the existing babel plugin is buggy and don't fit our purposes.

PR #59, opened 2017-10-19
fix land-on-master part of the autopublish travis script

fczuardi:
chore(autopublish): add semi columns to the branch = master case on the after success script (travis


fczuardi:
I dont know if this will fix, maybe there is a faster way to test it than opening multiple prs and then merging on master... I could start a autopublish-fakemaster branch to test it, I guess

fczuardi:
Ok, so, in order to make the life of the reviewers easier, I attempted the if case on another branch (master-sandbox) and it worked, see:

- https://travis-ci.org/CraveFood/farmblocks/builds/290113686#L826

and 

screen shot 2017-10-19 at 16 09 51

PR #60, opened 2017-10-20
Refactor/dismiss button

### LINKED ISSUE #56: [Alert] Use button component for dismiss button
Currently we use a styled button for the dismiss button.

We should use the farmblock button instead, to have consistency.

Also remove the "X" character by a proper icon font for the action.

### LINKED ISSUE #61: [Button] create a README with a png of the spec
Similar to the README we have on the Alert component

fczuardi:
Make use of the button farmblock inside the alert farmblock.

This PR ended up being a lot bigger than I've expected, because there was a necessary but unplanned work to be made on the Button component, that was different from the design spec, and also some work for using the local packages as dependencies on the development storybook server.

I am not quite happy with the current solution where jest expect the production version of the packages to run the tests and storybook expects the development version of the packages to bring up the local webpack without errors, I have opened issue #62 to target improvements on that part of the development workflow, but in summary here is the situation:

- if you want to locally start the storybook server, you have to first ``yarn build``
- and if you want to locally run the tests, you have to first ``BABEL_ENV=production yarn build``

This should be a temporary situation.

So, finally here is what this PR covers:

- [x] groundwork for a development environment that don't cause errors on storybook's webpack
- [x] closes #61 
- [x] design fixes to make button look like the spec
- [x] closes #56


packages/button/src/styledComponents/Button.js
```diff
@@ -48,7 +68,7 @@ function isIconOnly(props) {
 }
 
 function paddingStyle(props) {
-  const isLarge = props.size === LARGE;
+  const isLarge = props.size === MEDIUM;
```
alcferreira: please change `isLarge` to `isMedium`

fczuardi: feito

PR #63, opened 2017-10-25
[WIP] Feature/stateful alert

### LINKED ISSUE #28: [Alert] Implement a stateful version of the component with awareness of its own visibility
The current Alert component is a steteless "dumb" functional component and the responsibility of displaying it and hiding it is delegated to higher order components.

It may be useful to include in the package a version of the component that is able to hide itself independently of other components, either with a new property ``visibleTime`` or with a default onDismiss function that hides the Alert.

fczuardi:
- [ ] closes #28


fczuardi:
I am starting this branch with the dismiss-button branch as base with the hope that #60 will land on master. When that happens, I plan to rebase the stateful-alert branch to master.

PR #76, opened 2017-11-03
Stepper component

### LINKED ISSUE #68: Create onboard stepper component
Specs https://app.zeplin.io/project/595a9cd3b401bf1876faab27/screen/59f0de71521656e699a6c55b

image

### LINKED ISSUE #71: sync fonticon
the iconfont got updated with https://github.com/CraveFood/webapp/pull/2422 (added pork icon)

this issue is only about syncing the local dev version used by farmblock storybook assets.

alcferreira:
affects: @crave/farmblocks-stepper
Closes #68 
Closes #71


packages/stepper/src/components/Step.js
```diff
@@ -0,0 +1,55 @@
...
+
+  _renderIcon(iconName) {
+    return (
+      <div className="icon">
+        <i className={iconName} />
+      </div>
+    );
+  }
+
+  static propTypes = {
```
henlz: What do you think about putting the `propTypes` on the beginning of the file to make a faster reading of the components properties?

packages/stepper/src/styledComponents/Step.js
```diff
@@ -0,0 +1,52 @@
...
+import statusTypes from "../constants/statusTypes";
+
+const StepContainer = styled.div`
+  font-family: Lato, sans-serif;
+  size: 18px;
+  line-height: 1.33;
+
+  display: flex;
+
+  border-radius: 4px;
```
henlz: I think that it's a good idea to start thinking on a way to convert `pixel` values to `rem`

packages/stepper/src/components/Step.js
alcferreira: Hmm... On our webapp we always placed the `propTypes` at the end of the file.
I know it's not the same project, but i kinda like this way because when I open the file I can get faster to the `render` method which I believe is the most viewed piece of the code :stuck_out_tongue: 
Do you see any other advantages?

packages/stepper/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "files": ["AUTHORS", "LICENSE", "README.md", "CHANGELOG.md", "lib"],
+  "publishConfig": {
+    "access": "public"
+  },
+  "scripts": {
+    "test": "jest",
+    "clean": "rm -rf lib",
+    "build": "yarn clean && babel src -d lib --ignore '*.test.js,*.story.js'",
+    "prestorybook": "cp -Rf ../../.storybook .",
+    "storybook": "start-storybook -p 6006"
```
fczuardi: this is minor, as we rarely launch storybook from inside a package subfolder, but we should pass the assets I guess, replace this line with:

```
   "storybook": "start-storybook -p 6006 -s .storybook/assets",
```

packages/stepper/src/components/Step.js
```diff
@@ -0,0 +1,55 @@
...
+import statusTypes from "../constants/statusTypes";
+
+export default class Step extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.onClick = this.onClick.bind(this);
+  }
+
+  onClick(event) {
```
fczuardi: wouldnt it be better to do this wrapping on the parent component? and then just pass the onClick property to the current element?

fczuardi: one advantage of this wrapping belonging to the parent is that we can remove the need for an index property and make this step component simpler

packages/stepper/src/components/Step.js
```diff
@@ -0,0 +1,58 @@
...
+
+  render() {
+    const status = this.props.status;
+    const isCompleted = status === statusTypes.COMPLETED;
+
+    return (
+      <Container status={status} onClick={this.onClick}>
+        {this._renderIcon(isCompleted ? "wg-check" : "wg-checker")}
+
+        <div className="description">{this.props.value}</div>
```
fczuardi: what do you think about replacing the ``this.props.value`` property with ``this.props.children``? It might be useful in the future. (it would allow for "steps" arrays that can be nodes later, like to allow rich text)

packages/stepper/src/components/Step.js
```diff
@@ -0,0 +1,58 @@
...
+        <i className={iconName} />
+      </div>
+    );
+  }
+
+  static propTypes = {
+    index: PropTypes.number.isRequired,
+    value: PropTypes.string.isRequired,
+    status: PropTypes.oneOf(Object.keys(statusTypes)).isRequired,
+    onClick: PropTypes.func.isRequired
```
fczuardi: onClick could be optional if the wrapping was responsibility of the parent

packages/stepper/src/components/Step.js
henlz: For example, I developing another component and just wanna check the props of the Step component, it's way faster just to open the file and look at the first lanes then running a storybook or scrolling down to the render method.

packages/stepper/README.md
```diff
@@ -0,0 +1,63 @@
...
+## Usage
+
+Pass a list of `steps` to be completed, the quantity of `completedSteps` and a function to handle `onClick`.
+
+The `Stepper` component will render:
+  - `COMPLETED` status from index 0 of the `steps` array until the index of the last completed step (`completedSteps - 1`);
+  - `CURRENT` status to the next step after the last completed step
+  - `PENDING` status to all remaining steps
+
+`onClick` will be called only for `PENDING` steps.
```
fczuardi: CURRENT, right?

packages/stepper/src/constants/colorTypes.js
```diff
@@ -0,0 +1,21 @@
+import { colors as colorConstants } from "@crave/farmblocks-constants";
```
fczuardi: 👍

packages/stepper/README.md
alcferreira: Right! thanks

Issue #88, opened 2017-11-09
Centralize fluid button


alcferreira:
Fluid button text is aligned to the left. It should be centered.

Current:
![image](https://user-images.githubusercontent.com/17936244/32602975-4ed00954-c52f-11e7-96bb-2e8d2c56578f.png)

Specs: 
![image](https://user-images.githubusercontent.com/17936244/32602996-58652152-c52f-11e7-8abd-b947624689c9.png)

https://scene.zeplin.io/project/595a9cd3b401bf1876faab27/screen/59f0de71dcae46d5acf389ae

PR #97, opened 2017-11-09
Fix fluid button text alignment

fgameiro:
Closes: #88

### DIFF
--- a/__snapshots__/Storyshots.test.js.snap
+++ b/__snapshots__/Storyshots.test.js.snap
@@ -930,7 +930,7 @@ exports[`Storyshots Alert default 1`] = `
           className="sc-htpNat ihtkAy"
         >
           <button
-            className="sc-bxivhb iXegJq"
+            className="sc-bxivhb iCXejH"
             disabled={undefined}
             icon="wg-close"
             onClick={[Function]}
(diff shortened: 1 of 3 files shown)


fczuardi:
Ok, you need to update the snapshots with:

```
yarn test -u
```

fgameiro:
I was talking with @alcferreira about this: the tests are running fine locally -- 100% passed. But it's broken here on something I didn't even touch (alert components storyshots). We don't know why.

fczuardi:
I can reproduce it on my machine using ``yarn install --frozen-lockfile``

And the diff is for some classnames only, so it might be something on a dependency that changed the hash of the styled components of the alert component...

I will submit a commit and see if it fixes travis

fczuardi:
that didnt work, my local hash is still different from travis,looks like a bug on jest snapshots

fczuardi:
there is this project https://github.com/styled-components/jest-styled-components that can be a possible fix and makes styles changes more useful on snapshots as well, I am opening issue #106 to address that

fczuardi:
Ok, I found the cause of the different hashes... we should run the build task ``yarn build`` to generate the es5 version of the farmblocks components before generating our snapshots, because travis does that too.

PR #104, opened 2017-11-10
Text package

### LINKED ISSUE #103: Text component

alcferreira:
Closes #103


packages/text/src/components/Caption.js
```diff
@@ -13,8 +12,8 @@ export default class Caption extends React.Component {
 
   static propTypes = {
     children: PropTypes.node.isRequired,
-    size: PropTypes.oneOf(Object.keys(fontSizes)),
-    type: PropTypes.oneOf(Object.keys(fontTypes)),
+    size: PropTypes.oneOf(Object.values(fontSizes)),
```
fczuardi: I think we can leave this property type flexible, and use the fontSizes just as suggestion

packages/theme/src/fontSizes.js
```diff
@@ -1,20 +1,7 @@
...
-export default {
-  keys,
-  values
+module.exports = {
+  HUGE: 24,
+  X_LARGE: 20,
+  LARGE: 18,
+  MEDIUM: 16,
+  SMALL: 14
 };
```
fczuardi: 👍

packages/text/README.md
```diff
@@ -0,0 +1,37 @@
+![logo-farmblocks](https://user-images.githubusercontent.com/7760/31051341-4d280118-a63c-11e7-9e8f-3b375ca8f9a0.png)
+
+# Farmblocks-Theme
```
fczuardi: Farmblocks-Text

packages/text/README.md
```diff
@@ -0,0 +1,37 @@
...
+## Usage
+
+```jsx
+import React from 'react';
+import { render } from 'react-dom';
+import { Title, Caption, Paragraph, fontSizes, fontTypes } from '@crave/farmblocks-text';
+
+render(
+  <div>
+    <Title size={fontSizes.X_LARGE} type={fontTypes.NORMAL}>Content title </Title>
```
fczuardi: I was thinking about EXTRA_LARGE vs X_LARGE...

On the Image component, we went with EXTRA, but X gives us the advantage of being able to use XX_SMALL and XXX_SMALL in the future, similar to the css api...

The other argument would be that size names are more semantic and more than 5 variations might be too much, so TINY and HUGE as aliases for XX on each direction can be enough, and if the need for XXX appears we could use EXTRA_TINY and EXTRA_HUGE

I personally dont have a strong preference on either choice, but it would be nice to use the same choice for different components, for consistence sake

packages/text/src/components/Caption.js
```diff
@@ -0,0 +1,25 @@
...
+
+export default class Caption extends React.Component {
+  render() {
+    const { children, ...props } = this.props;
+    return <Container {...props}>{children}</Container>;
+  }
+
+  static propTypes = {
+    children: PropTypes.node.isRequired,
+    size: PropTypes.oneOf(Object.values(fontSizes)),
```
fczuardi: I think we can leave sizes flexible instead of restricted to our sizes.

packages/text/src/styledComponents/Caption.js
```diff
@@ -0,0 +1,14 @@
+import styled from "styled-components";
+
+const CaptionContainer = styled.div`
+  font-family: Lato, sans-serif;
```
fczuardi: Maybe we should include line-height here (and on the others), Zeplin shows different line heights for the different sizes

packages/text/README.md
alcferreira: We went with the X_LARGE option.

fgameiro:
Looks good to me.

Issue #78, opened 2017-11-07
Content badge component


alcferreira:
https://scene.zeplin.io/project/595a9cd3b401bf1876faab27/screen/59f0df31bbc93198a853bbe2
https://scene.zeplin.io/project/595a9cd3b401bf1876faab27/screen/59f0df2fee4fbc85eccb17b1

PR #108, opened 2017-11-13
content badge component (Image)

fczuardi:
Closes: #78

### DIFF
--- a/packages/image/README.md
+++ b/packages/image/README.md
@@ -17,7 +17,7 @@ TBD
 const React = require("react");
 const ReactDOM = require("react-dom");
 const Image = require("@crave/farmblocks-image").default;
-const { thumbnailSizes, badgeSizes } = require("@crave/farmblocks-image")";
+const { imageTypes, thumbnailSizes, badgeSizes } = require("@crave/farmblocks-image")";
 
 const root = document.createElement("div");
 
--- a/packages/image/README.md
+++ b/packages/image/README.md
@@ -9,7 +9,8 @@ npm install @crave/farmblocks-image
 ```
 
 ## Spec
-TBD
+Thumbnails: https://scene.zeplin.io/project/595a9cd3b401bf1876faab27/screen/59f0df2fee4fbc85eccb17b1
+Badges: https://scene.zeplin.io/project/595a9cd3b401bf1876faab27/screen/59f0df31bbc93198a853bbe2
 
 ## Usage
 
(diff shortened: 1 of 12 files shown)


packages/image/src/Image.story.js
```diff
@@ -3,110 +3,85 @@ import { storiesOf } from "@storybook/react";
...
 const imgSrc =
   "https://sourcewhatsgood.com/assets/images/utility_images/gallery-third-about-a518a29f64.jpg";
 
 storiesOf("Image", module)
-  .add("default", () => <Image src={imgSrc} />)
-  .add("thumbnail", () => <Image src={imgSrc} type={imageTypes.THUMBNAIL} />)
-  .add("badge", () => <Image src={imgSrc} type={imageTypes.BADGE} />)
+  .add("default", withInfo()(() => <Image src={imgSrc} />))
+  .add("thumbnail", withInfo()(() => <Image src={imgSrc} />))
+  .add("badge", withInfo()(() => <Image src={imgSrc} badge={true} />))
```
alcferreira: When a property is a boolean, you don't need to explicitly set it to true.

You can write it like this: `<Image src={imgSrc} badge />`. If you don't pass the prop, it will be false.

Please refactor all occurrences of `badge={true}`;

fgameiro: It's done.

packages/image/README.md
```diff
@@ -0,0 +1,42 @@
...
+A react component for thumbnail images
+
+## Installation
+
+```
+npm install @crave/farmblocks-image
+```
+
+## Spec
+TBD
```
fczuardi: We should link zeplin here

packages/image/README.md
```diff
@@ -0,0 +1,42 @@
...
+const ReactDOM = require("react-dom");
+const Image = require("@crave/farmblocks-image").default;
+const { imageTypes, thumbnailSizes, badgeSizes } = require("@crave/farmblocks-image")";
+
+const root = document.createElement("div");
+
+ReactDOM.render(
+  <Image
+    src=""
+    type={imageTypes.BADGE}
```
fczuardi: We should update this snippet and test it on https://stackblitz.com

packages/image/src/styledComponents/Image.js
```diff
@@ -0,0 +1,30 @@
+import styled from "styled-components";
+import thumbnailSizes from "../constants/thumbnailSizes";
+import { colors as colorConstants } from "@crave/farmblocks-theme";
```
fczuardi: We need to add this package as a dependency.

PR #110, opened 2017-11-14
fix(button): fix fluid style

alcferreira:
affects: @crave/farmblocks-button

ISSUES CLOSED: #109

### DIFF
--- a/packages/button/src/styledComponents/Container.js
+++ b/packages/button/src/styledComponents/Container.js
@@ -3,7 +3,6 @@ import styled from "styled-components";
 
 const Container = styled.div`
   position: relative;
-  display: ${props => (props.fluid ? "block" : "inline-block")};
 
   > .hit-box-container {
     position: absolute;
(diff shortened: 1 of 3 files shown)


packages/button/src/styledComponents/Container.js
```diff
@@ -1,16 +1,27 @@
...
   }
+
+  ${fluidStyle};
 `;
 
+function fluidStyle(props) {
+  return (
+    props.fluid &&
+    css`
+      max-width: 100%;
```
fczuardi: is max-width necessary?

packages/button/src/styledComponents/Container.js
```diff
@@ -1,16 +1,27 @@
 // @flow
-import styled from "styled-components";
+import styled, { css } from "styled-components";
 
 const Container = styled.div`
   position: relative;
-  display: ${props => (props.fluid ? "block" : "inline-block")};
```
fczuardi: is the default display value (inline-block) set somewhere else?

alcferreira: don't think so... is it necessary anywhere?

packages/button/src/styledComponents/Container.js
alcferreira: Not really. i'll remove

PR #113, opened 2017-11-20
travis is erroring when landing on master

fczuardi:
after greenkeeper merges we need to update the lockfile
:(


fczuardi:
this looks like the same issue as https://github.com/greenkeeperio/greenkeeper-lockfile/issues/85

I will open an issue for us to workaround it if possible

Issue #112, opened 2017-11-17
Link component


alcferreira:
Create link component https://scene.zeplin.io/project/595a9cd3b401bf1876faab27/screen/59ee42d9b95cec4ea9ed4e14

PR #117, opened 2017-11-20
Feat: Link component

fgameiro:
affects: @crave/farmblocks-link

Closes:
#112


yarn.lock
```diff
@@ -1,10 +1,24 @@
 # THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
fczuardi: you could do a rebase with master instead of a merge, and if you do a ``git rebase -i master`` you can group and squash similar commits as well, such as the "snapshot update" ones

.gitignore
```diff
@@ -16,5 +16,14 @@ npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 
+<<<<<<< HEAD
```
fczuardi: looks like leftover from conflict resolution, clean those

.storybook/assets/fonticon.scss
```diff
@@ -0,0 +1,377 @@
+@font-face {
```
fczuardi: there is no need for this scss file, please remove it

package.json
```diff
@@ -72,6 +72,7 @@
     "jest-styled-components": "^4.9.0",
     "lerna": "^2.4.0",
     "lint-staged": "^5.0.0",
+    "pre-commit": "^1.2.2",
```
fczuardi: we use the python pre-commit, not the npm one, please remove this line

package.json
```diff
@@ -37,7 +37,7 @@
     "url": "git+https://github.com/CraveFood/farmblocks.git"
   },
   "devDependencies": {
-    "@crave/farmblocks-tooltip": "^0.2.0-alpha.e320de4a",
+    "@crave/farmblocks-tooltip": "^0.3.0",
```
fczuardi: I think it is safe to remove this line as well, since tooltip is on the workspaces already

packages/link/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+```jsx
+import React from "react";
+import ReactDOM from "react-dom";
+import Link from "@crave/farmblocks-link";
+
+const root = document.createElement("div");
+
+ReactDOM.render(
+  <Link>This is a link!</Link>,
```
fczuardi: href is a required attribute

packages/link/src/constants/linkTypes.js
```diff
@@ -0,0 +1,10 @@
+import { fontTypes } from "@crave/farmblocks-theme";
+
+const linkTypes = {
```
fczuardi: maybe this file should be named linkColors, since the values on this dict are color hex values in the end of the day

fgameiro: I named it this way because of the fontTypes import, which is also a color dictionary, so using it in a 'link' context became 'linkTypes'

yarn.lock
fgameiro: I wasn't sure what I was doing when I had to include the most recents commits from master (containing the tooltip fixes) to my working branch, so I was afraid of messing up the commits history. Should I do a rebase now?

packages/link/src/constants/linkTypes.js
fczuardi: ok, not a big deal

packages/link/src/Link.js
```diff
@@ -0,0 +1,84 @@
...
+  static propTypes = {
+    children: PropTypes.node.isRequired,
+    href: PropTypes.string.isRequired,
+    type: PropTypes.oneOf(Object.values(linkTypes)),
+    tooltipAlign: PropTypes.oneOf(["left", "right"]),
+    leftIcon: PropTypes.string,
+    disabled: PropTypes.bool,
+    external: PropTypes.bool,
+    size: PropTypes.number,
+    lineHeight: PropTypes.number
```
fczuardi: the Text component accepts both number or string for lineHeight, we should copy that here

packages/link/src/Link.js
```diff
@@ -0,0 +1,84 @@
...
+    this.mouseOver = this.mouseOver.bind(this);
+    this.mouseLeaves = this.mouseLeaves.bind(this);
+  }
+
+  render() {
+    const { children, ...props } = this.props;
+    return (
+      <Container {...props}>
+        {props.disabled && (
+          <div
```
fczuardi: This div feels hackish.
Maybe remove this hit box and replace the ``<a>`` tag with a ``<span>`` tag when disabled, and attach the over/leave events to that tag.
By changing the markup we have a cleaner alternative to the current css-based way of preventing a click.

packages/link/src/styledComponents/Link.js
```diff
@@ -0,0 +1,78 @@
...
+import linkTypes from "../constants/linkTypes";
+
+const LinkContainer = styled.div`
+  font-family: Lato, sans-serif;
+
+  display: inline-block;
+
+  position: relative;
+
+  > .hit-box-container {
```
fczuardi: see the discussion about just changing the a tag to text

packages/link/src/styledComponents/Link.js
```diff
@@ -0,0 +1,78 @@
...
+function textDecoration({ disabled, type }) {
+  const underline =
+    disabled || type === linkTypes.NEUTRAL || type === linkTypes.WHITE
+      ? "none"
+      : "underline";
+
+  return css`
+    & > a {
+      text-decoration: none;
+      pointer-events: ${disabled && "none"};
```
fczuardi: see the discussion about just changing the a tag to text

packages/link/src/Link.js
fgameiro: Sure, this looks much better

packages/link/src/Link.js
```diff
@@ -0,0 +1,84 @@
...
+    if (this.props.disabled) {
+      this.setState({
+        showTooltip: false
+      });
+    }
+  }
+
+  static propTypes = {
+    children: PropTypes.node.isRequired,
+    href: PropTypes.string.isRequired,
```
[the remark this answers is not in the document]
fgameiro: I don't know... right now I wouldn't know what are the chances that we may need it... @alcferreira what do you think?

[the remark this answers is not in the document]
alcferreira: Hmmm don't think so right now...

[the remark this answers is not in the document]
alcferreira: Actually we`ll need this

packages/link/src/Link.js
```diff
@@ -0,0 +1,82 @@
...
+            {children}
+          </span>
+        ) : (
+          <a href={props.href}>{children}</a>
+        )}
+
+        {props.external && <i className="wg-external-link margin-left" />}
+
+        <Tooltip
+          text="This action is disabled."
```
fczuardi: can we turn this into a property? it would help with localization. Having hardcoded text is not a good idea, I have filed issue #135 to change this on the button  component as well

packages/link/src/Link.Neutral.story.js
```diff
@@ -0,0 +1,50 @@
...
+    withInfo()(() => (
+      <Link href="#" type={linkTypes.NEUTRAL}>
+        Awesome Link
+      </Link>
+    ))
+  )
+  .add(
+    "right aligned",
+    withInfo()(() => (
+      <div style={{ width: "500px", textAlign: "right" }}>
```
fczuardi: when a value is in pixels, React allows it to be set as number, just a tip, not a request

fczuardi:
👍

fczuardi:
@fgameiro can you please rebase this with the current master, see if the snapshots change and push it again please?

Issue #121, opened 2017-11-22
Use Link component on Footer


alcferreira:
- Needs: #112

PR #136, opened 2017-11-30
refactor(Footer): add farmblocks-link to Footer component

alcferreira:
affects: @crave/farmblocks-footer

ISSUES CLOSED: #121

### DIFF
--- a/packages/footer/src/components/Footer.js
+++ b/packages/footer/src/components/Footer.js
@@ -13,7 +13,7 @@ const Footer = props => {
       linkText={props.helpLinkText}
       linkHref={props.helpLinkHref}
       imageSrc={props.helpImageSrc}
-      onClick={props.helpLinkOnClick}
+      linkOnClick={props.helpLinkOnClick}
     />
   );
   if (props.noAction) {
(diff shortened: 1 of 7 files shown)


packages/footer/src/components/Footer.js
```diff
@@ -52,6 +53,7 @@ Footer.propTypes = {
   helpText: PropTypes.string.isRequired,
   helpLinkText: PropTypes.string,
   helpLinkHref: PropTypes.string,
+  helpLinkOnClick: PropTypes.func,
```
fczuardi: maybe this name can be ``onHelpLinkClick`` to follow the others: ``onSecondaryActionClick`` and ``onPrimaryActionClick``

packages/footer/src/components/HelperContent.js
```diff
@@ -37,7 +42,8 @@ HelperContent.propTypes = {
   imageSrc: PropTypes.string,
   text: PropTypes.string.isRequired,
   linkText: PropTypes.string,
-  linkHref: PropTypes.string
+  linkHref: PropTypes.string,
+  linkOnClick: PropTypes.func
```
fczuardi: maybe the above comment applies here as well... onLinkClick?

packages/footer/src/components/Footer.js
alcferreira: I came up with `helpLinkOnClick` so keep the prefix of the other props. But sure, i'll change that

PR #148, opened 2017-12-04
Text Input component

### LINKED ISSUE #80: Input component
https://scene.zeplin.io/project/595a9cd3b401bf1876faab27/screen/59ee4500c5418ce9dc281888

fczuardi:
affects: ``@crave/farmblocks-input-text``
- closes #80


packages/input-text/src/TextInput.js
```diff
@@ -0,0 +1,165 @@
...
+        <div className="messages">
+          {errors.map((text, index) => (
+            <div className="message" key={`err-${index}`}>
+              <Image
+                className="icon"
+                src={iconSrc}
+                badge
+                size={badgeSizes.SMALL}
+              />
+              {text}
```
alcferreira: Image and text are misaligned...
It should be vertically center aligned.

Also use `@crave/farmblocks-text`component to render all texts. This way we can follow a pattern and centralize the text styling.

image

packages/input-text/src/styledComponents/TextInput.js
```diff
@@ -0,0 +1,85 @@
...
+  flex-direction: column;
+  margin-bottom: 24px;
+
+  .label {
+    order: 1;
+    text-transform: uppercase;
+    font-size: ${fontSizes.SMALL}px;
+    margin-bottom: 8px;
+    color: ${labelColor};
+  }
```
alcferreira: Missing `font-weight`, `line-height` and `letter-spacing`.

image

packages/input-text/src/TextInput.js
```diff
@@ -0,0 +1,165 @@
...
+      onChange: this.onChange,
+      onInvalid: this.onInvalid,
+      onMouseOver: this.onMouseOver,
+      onMouseLeave: this.onMouseLeave,
+      ...otherProps
+    };
+    return (
+      <StyledLabel {...containerProps}>
+        {this._renderInput(inputProps)}
+        {this._renderLabel(label)}
```
alcferreira: Any special reason to render input before `label` after `input` and change it with `order` css?

packages/input-text/src/TextInput.js
```diff
@@ -0,0 +1,165 @@
...
+            </div>
+          ))}
+        </div>
+      )
+    );
+  }
+  onChange(event) {
+    this.setState({
+      value: event.target.value,
+      invalid: false
```
alcferreira: Not sure about this... This will work with the `required` prop and with other validations... (?)

packages/input-text/src/styledComponents/TextInput.js
fczuardi: oops, good catch

packages/input-text/src/TextInput.js
```diff
@@ -0,0 +1,165 @@
...
+    });
+    return this.props.onChange(event);
+  }
+
+  onInvalid(event) {
+    event.preventDefault();
+    this.setState({ invalid: true });
```
alcferreira: We must keep in mind that the invalid state of the component will be handled by the wg-api (on our case).
This way we must explicitly control the `invalid` prop with new props passed.

What happens if a field is mounted with valid status, then we make some changes and make an api call, this api call returns an error. How do we handle this situation?

packages/input-text/src/TextInput.js
fczuardi: to save the use of the for property and generate unique ids with the label after the input we can switch to the focused color with css only

packages/input-text/src/TextInput.js
```diff
@@ -0,0 +1,165 @@
...
+
+  static defaultProps = {
+    value: "",
+    invalid: false,
+    onChange: () => null,
+    onInvalid: () => null,
+    onMouseOver: () => null,
+    onMouseLeave: () => null,
+    errorIconSrc,
+    validationErrors: ["This field is required"],
```
alcferreira: I don't think we should have a `default` validation message... 
What if we controll the `invalid` state with this field? I mean: if we have `validationErrors.length` > 0, field is invalid.

packages/input-text/src/TextInput.js
fczuardi: this version of the component dont have support for custom validations, and I don;t know if the DOM provides hooks to the native validations, I can investigate it. Yes, this hardcoded simple "have changed = valid now" is not ideal

packages/input-text/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Installation
+
+```
+npm install @crave/farmblocks-input-text
+```
+
+## License
+
+MIT
```
alcferreira: todo: write readme

packages/input-text/src/TextInput.js
alcferreira: didn't find `for` on label... But ok, thanks

packages/input-text/src/TextInput.js
fczuardi: there is a boolean invalid property, webapp can change that property to true upon receiving the errors from the backend

fczuardi: this default wrrapper is for the component to be able to flag as invalid empty fields before submitting (html 5 required attribute)

fczuardi: and to override the default native balloon from the browsers

packages/input-text/src/TextInput.js
fczuardi: ok, let me rethink this api then... 

having no default messages, would make the native bubble that I am preventing default behavior less useful, and I think it would be nice to mirror html's form validation interface. I will try to purpose something better, and if I fail I can implement your proposed implicit api 
(no messages array = no validation erros)

packages/input-text/src/TextInput.js
fczuardi: yes, I am not using `for` 

in my comment above, I've use "to save" meaning "to avoid"

packages/input-text/src/TextInput.js
fczuardi: I have fixed this alignment by setting the text line height to be the same of the image

fczuardi: I have added farmblocks-text to the dependencies and have used for label and for the error message, but I still had to add some rules on top of that, like the letter-spacing from your other request

packages/input-text/src/styledComponents/TextInput.js
fczuardi: font-weight 600 was possible using the title property of the text component, line height and letter spacing I had to set

packages/input-text/src/TextInput.js
fczuardi: I have added a comment on the code explaining this choice, but if you think a different implementation is better let me know

packages/input-text/src/TextInput.js
fczuardi: by other validations do you mean https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/HTML5/Constraint_validation ? 

I only tested with ``required`` (the "missing" validation) I haven't covered the other cases like type, pattern, underflow, overflow, step and too long

I will take a look at the link above and also this interface https://developer.mozilla.org/en-US/docs/Web/API/ValidityState and see what can be done to this milestone

packages/input-text/README.md
fczuardi: this document will be real useful here, to describe which properties and apis we mirror from html and which ones differs or are unimplemented, will write this

packages/input-text/src/TextInput.js
fczuardi: a better value for this invalid flag would be ``event.target.validity.valid`` that is the property when using the built-in validation

alcferreira: Nice!

packages/input-text/src/TextInput.js
alcferreira: Here's the situation that I've commented above
https://stackblitz.com/edit/react-z67acp?file=index.js

packages/input-text/src/styledComponents/TextInput.js
```diff
@@ -0,0 +1,85 @@
...
+const inputBackgroundColor = props => {
+  return props.disabled ? colors.GREY_16 : "white";
+};
+const inputBoxShadow = focused => props => {
+  if (!focused && (props.disabled || props.filled)) {
+    return css`
+      box-shadow: none;
+    `;
+  }
+  const shadowSize = props.invalid ? "0 2px 2px 0" : "0 4px 4px 0";
```
alcferreira: shadowSize is `4px` if the field is focused otherwise it's `2px`

packages/input-text/src/styledComponents/TextInput.js
```diff
@@ -0,0 +1,85 @@
...
+  margin-bottom: 24px;
+
+  .label {
+    order: 1;
+    text-transform: uppercase;
+    letter-spacing: 2px;
+    margin-bottom: 8px;
+    color: ${labelColor};
+  }
+  input:focus + .label {
```
alcferreira: Good!

PR #168, opened 2017-12-14
HOC disabled tooltip

### LINKED ISSUE #166: Abstract the behavior of displaying tooltip for disable components into a HOC
Currently 3 components have a very similar behaviour:

- a ``disabled`` property
- that when true sets the onMouseOver and onMouseLeave event listeners
- to display and hide a tooltip
- that can have it's default text changed using a ``tooltipText`` property

We should standardise it and centralize it in a lib imported by all the blocks with this behaviour and use it as a decorator.

fczuardi:
- closes #166


packages/hoc-disabled-tooltip/src/disabledTooltip.test.js
```diff
@@ -0,0 +1,83 @@
...
+    wrappedComponent.simulate("mouseover", {});
+    wrappedComponent.simulate("mouseleave", {});
+    expect(onMouseLeaveMock).toBeCalled();
+    expect(onMouseOverMock).toBeCalled();
+    expect(onClickMock).toBeCalled();
+  });
+
+  test("mouseover and mouseleave events should change showTooltip state on disabled fields", function() {
+    const EnhancedInput = disabledTooltip("input");
+    let component = mount(<EnhancedInput disabled />);
```
alcferreira: Use `const` instead

packages/hoc-disabled-tooltip/README.md
```diff
@@ -0,0 +1,20 @@
...
+```
+
+## Usage
+
+```javascript
+```
+
+See more examples in the stories file.
+
+## License
```
alcferreira: Empty readme 😅

packages/hoc-disabled-tooltip/src/disabledTooltip.story.js
```diff
@@ -0,0 +1,49 @@
...
+  .add(
+    "disabled html input with custom Tooltip text",
+    withInfo()(() => (
+      <EnhancedInput
+        type="checkbox"
+        disabled
+        tooltipText="A custom tooltip text"
+      />
+    ))
+  );
```
alcferreira: I think it'd be nice to have an example passing a React component to the hoc instead of only passing the `type` as string. I get that the output is the same, but for documentation sake (:

fczuardi: The usage on the Readme has such an example now, but I can add one to the stories as well...

fczuardi:
comments addressed in the latest 4 commits

packages/hoc-disabled-tooltip/src/styledComponents/Wrapper.js
```diff
@@ -0,0 +1,18 @@
+import styled from "styled-components";
+
+const Wrapper = styled.div`
+  display: inline-block;
```
fczuardi: this can break the ``fluid`` option on the button component

fczuardi: The reason this inline block exists:

Not all form inputs can trigger the onMouseOver and onMouseLeave events, ex: a disabled checkbox input, buttons, etc. So one way to display tooltip on mouseover is to have a hit area overlay on top of the wrapped component. And since we dont know the width of the element, we use a container with inline style to host both the wrapped component and t he hit area.

Also, the main use of this hoc is for inline elements such as links, buttons and form inputs...

But this can break the expected layout widths of block elements witth width set in percents, like a button with 100% width (the case of a farmblocks button with the fluid flag set).

I dont know of an elegant way to handle that case, one simple solution was to have a new property to cover this exception and remove the default display inline behaviour when needed.

For this moment I am submitting a patch that implements a new ``displayBlock`` property. I am not super fan of this idea and I am open to suggestions, either on a separate issue or on this same thread.

fczuardi:
implemented the patch [described here](https://github.com/CraveFood/farmblocks/pull/168#discussion_r159408991)

packages/hoc-disabled-tooltip/src/styledComponents/Wrapper.js
fczuardi: the text-input previous default was display block, so I am thinking in reversing this again, make display block be the default, remove the displayBlock property and leave it for the user to wrap it in another div if she needs specific widths.

This will be included in the https://github.com/CraveFood/farmblocks/pull/171 patch

fczuardi: instead of reversing, I ended up just adding the displayBlock in the properties internally on th e input-text refactorying...

PR #177, opened 2017-12-19
Hoc input

### LINKED ISSUE #165: abstract common form input properties / behavior / styles into a higher order component
After the input text component, we will need other form inputs, like checkbox, radio buttons, select boxes, textarea, etc... and those components will share several characteristics like:

- bigger drop shadow when unfilled > smaller when focused > no after filled
- different border colors for unfocused, focused and errored
- prevent default browser balloon on the invalid event and wrap the event handler on our custom wrapper
- handle the value of the input in the state of the component (controlled component)
- ...

This issue is about creating a new generic component with those common features do be used as a dependency in the text input and other future input components

fczuardi:
- closes #165


packages/hoc-input/src/formInput.js
```diff
@@ -0,0 +1,175 @@
...
+      );
+    }
+    onChange(event) {
+      const validationMessages =
+        this.props.validationMessages.length > 0
+          ? this.props.validationMessages
+          : [];
+
+      this.setState({
+        value: event.target.value,
```
alcferreira: Got this error using this HOC on a stripe component:

image

Stripes calls `onChange` passing this object:
image

See reference on https://stripe.com/docs/stripe-js/reference#element-on

Issue #184, opened 2017-12-26
Dont update value based on props if it's not specified


alcferreira:
`hoc-input` must control the value either by 
- receiving value from props
- setting state

Currently if the component does not receive a `value` property, the value gets reseted everytime it receives new props.

PR #185, opened 2017-12-26
fix(hoc-input): do not update the state value if it does not come from properties

alcferreira:
affects: @crave/farmblocks-hoc-input

ISSUES CLOSED: #184


alcferreira:
@fczuardi me too. I'm changing one test that covers this to me more reliable. One sec.

fczuardi:
👍

Issue #197, opened 2018-01-08
Refactor our uses of Object.values


fczuardi:
This is the farmblocks side of issue https://github.com/CraveFood/webapp/issues/2570

Basically, old Safaris and IE don't support Object.values and this is something we are using on the proptypes of link and text components.

PR #202, opened 2018-01-08
fix(compatibility): use a shim for Object.values

fczuardi:
This can be a little tricky to test, you will need to find a browser that don't support Object.values to compare the behavior before and after the patch. I will comment some directions if I find simple ways of doing that, and having some process for us to test different browsers is a discussion/enhancement that we need for farmblocks and webapp, maybe we could file an issue for that.

affects: @crave/farmblocks-link, @crave/farmblocks-text

ISSUES CLOSED: #197


fczuardi:
rebased and conflicts resolved

fczuardi:
also, browserstack invites were sent

vNakamura:
😓 Storybook already uses a polyfill for Object.values. So we need to find another way to test it.

fczuardi:
@vNakamura https://stackblitz.com with the README usage code of the affected components?

first with the regular version, to see the failing, then with the specific canary of this branch (need to copy paste from travis log) to see the fix?

vNakamura:
@fczuardi Great, I could test this way!
To other reviewers, the version is: `@crave/farmblocks-input-text@0.3.0-alpha.a4bbf04a`

vNakamura:
Coverage reduced

fczuardi:
rebased

fczuardi:
rebased and resolved conflicts

Issue #208, opened 2018-01-12
Add rightIcon property to Button component


alcferreira:
We need a icon aligned to the right to use on Dropdown component

PR #209, opened 2018-01-12
feat(button): add rightIcon prop to Button

vNakamura:
affects: @crave/farmblocks-button

ISSUES CLOSED: #208

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -93,6 +93,10 @@ injectGlobal`
     content: "\\e973";
   }
 
+  .wg-small-arrow-bottom:before {
+    content: "\\e942";
+  }
+
   .wg-order,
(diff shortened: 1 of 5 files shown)


alcferreira:
Nice!
Just another improvements:

------------------
- [ ] When the button has no text at all, we are getting an extra margin after the left icon and before the right icon. Please insert this margin only if there is context on button.

this is the left icon image

this is the right icon image

------------------
- [ ] The icon is not vertically aligned.
Possible solution: add `vertical-align: text-bottom;` to it.

before 
image

after
image

------------------
- [ ] Fix margin after icon when passing children to `Button`.
e.g. `<Button icon="wg-order">{this.props.text}</Button>` gets rendered to:
image

Issue #201, opened 2018-01-08
Text filter component


alcferreira:
https://app.zeplin.io/project/595a9cd3b401bf1876faab27/screen/59f0dea1176ce3b4a024ca0c

![image](https://user-images.githubusercontent.com/17936244/34690520-a10e1120-f4a0-11e7-8427-059962fa1226.png)

PR #219, opened 2018-01-18
Filter Text Input

vNakamura:
affects: `@crave/farmblocks-input-text`

ISSUES CLOSED: #201, #224

### DIFF
--- a/packages/hoc-input/src/formInput.js
+++ b/packages/hoc-input/src/formInput.js
@@ -12,6 +12,7 @@ import Wrapper from "./styledComponents/Wrapper";
 export const formInputProps = {
   label: PropTypes.string,
   value: PropTypes.string,
+  type: PropTypes.string,
   focused: PropTypes.bool,
   disabled: PropTypes.bool,
   validationMessages: PropTypes.arrayOf(PropTypes.string),
@@ -19,12 +20,7 @@ export const formInputProps = {
   onChange: PropTypes.func,
   onInvalid: PropTypes.func,
   onFocus: PropTypes.func,
-  onBlur: PropTypes.func,
-  compact: PropTypes.bool
-};
-
-const searchProps = {
-  icon: "wg-search"
+  onBlur: PropTypes.func
 };
 
 const formInput = WrappedComponent => {
@@ -55,7 +51,6 @@ const formInput = WrappedComponent => {
         onFocus,
         onBlur,
         onInvalid,
-        compact,
         ...wrappedComponentProps
       } = this.props;
       const wrapperProps = {
@@ -63,7 +58,7 @@ const formInput = WrappedComponent => {
         invalid: this.state.validationMessages.length > 0,
         filled: this.state.value.length > 0,
         disabled: wrappedComponentProps.disabled,
-        compact
+        type: wrappedComponentProps.type
       };
       return (
         <Wrapper {...wrapperProps} onClick={this.handleWrapperClick}>
@@ -84,12 +79,9 @@ const formInput = WrappedComponent => {
 
       const isSearch =
         inputProps.type && inputProps.type.toLowerCase() === "search";
-      const composedProps = isSearch
-        ? { ...searchProps, ...inputProps }
-        : inputProps;
-      const icon = composedProps.icon && (
+      const icon = isSearch && (
         <div className="icon">
-          <i className={composedProps.icon} />
+          <i className="wg-search" />
         </div>
       );
       const clearButton = isSearch &&
@@ -108,7 +100,7 @@ const formInput = WrappedComponent => {
         >
           {icon}
           <WrappedComponent
-            {...composedProps}
+            {...inputProps}
             {...handlers}
             value={this.state.value}
           />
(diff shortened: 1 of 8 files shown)


packages/input-text/src/TextInput.js
```diff
@@ -5,15 +5,37 @@ import disabledTooltip, {
...
-  </Container>
-);
+const searchProps = {
+  icon: "wg-search",
+  clearable: true
+};
+
+const TextInput = props => {
+  if (props.type.toLowerCase() === "search")
+    props = { ...searchProps, ...props };
```
alcferreira: not cool to mutate a parameter 😢 
We can add a eslint rule to check this situation and warn us (:

packages/input-text/src/TextInput.js
```diff
@@ -5,15 +5,37 @@ import disabledTooltip, {
...
 const EnhancedInput = compose(disabledTooltip, formInput)("input");
 
-const TextInput = props => (
-  <Container>
-    <EnhancedInput {...props} displayBlock />
-  </Container>
-);
+const searchProps = {
+  icon: "wg-search",
+  clearable: true
```
fczuardi: do we need clearable? dont all type=search inputs have this feature by default? how is it on our specs?

fczuardi: allsearch fields have the clear button, see slack discussion https://cravefoodservices.slack.com/archives/C1AB7U58W/p1516300464000496

vNakamura: I though on having `clearable` so it could be used by other types of inputs.
But I we can do this only when/if we need.

packages/hoc-input/src/formInput.js
```diff
@@ -73,13 +78,35 @@ const formInput = WrappedComponent => {
...
         <div className="input">
+          {icon}
           <WrappedComponent
-            {...inputProps}
+            {...composedProps}
             {...handlers}
             value={this.state.value}
+            ref={input => {
+              this.wrappedComponent = input;
+            }}
```
vNakamura: I don't understand why ref is not working.
I'm getting this: `Warning: Stateless function components cannot be given refs. Attempts to access this ref will fail.` but this is not a stateless function.

fczuardi:
on Chrome, both cancel buttons appear:

screen shot 2018-01-22 at 14 24 11

This can either be styled, or hidden on webkit/blink based browsers using the non-standard pseudoclass [::-webkit-search-cancel-button](https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-search-cancel-button)

packages/hoc-input/src/formInput.js
```diff
@@ -18,7 +19,12 @@ export const formInputProps = {
   onChange: PropTypes.func,
   onInvalid: PropTypes.func,
   onFocus: PropTypes.func,
-  onBlur: PropTypes.func
+  onBlur: PropTypes.func,
+  compact: PropTypes.bool
+};
+
+const searchProps = {
+  icon: "wg-search"
```
fczuardi: I think this can go inline in the icon markup only

packages/hoc-input/src/formInput.js
```diff
@@ -73,13 +81,38 @@ const formInput = WrappedComponent => {
         onFocus: this.onFocus,
         onBlur: this.onBlur
       };
+
+      const isSearch =
+        inputProps.type && inputProps.type.toLowerCase() === "search";
+      const composedProps = isSearch
```
fczuardi: same as the previous comment, we can get rid of this variable

packages/hoc-input/src/formInput.js
```diff
@@ -135,6 +168,19 @@ const formInput = WrappedComponent => {
       this.setState(nextState);
     }
 
+    handleClearClick = () => {
+      this.setState({ value: "" });
+      this.props.onChange({
+        type: "change",
+        value: "",
+        target: { value: "" }
```
fczuardi: this should be fine, but passing target: this.inputRef should also work, right?

packages/hoc-input/src/formInput.story.js
```diff
@@ -73,4 +73,17 @@ storiesOf("HOC formInput", "module")
         <EnhancedInput disabled label="Input Label" type="text" />
       </div>
     ))
+  )
+  .add(
+    "html text input with icon",
```
fczuardi: Do we have text inputs with icon that is not search?

packages/hoc-input/src/formInput.test.js
```diff
@@ -123,4 +124,22 @@ describe("formInput", function() {
...
+    const focusMock = jest.fn();
+    const component = mount(<EnhancedInput />);
+    const wrapper = component.find(Wrapper);
+    const input = component.find("input").instance();
+    input.focus = focusMock;
+    wrapper.simulate("click");
+    expect(focusMock).toHaveBeenCalledTimes(1);
+  });
+
+  test("get have no value when clear button is clicked", function() {
```
fczuardi: minor text change, the name of this test is strange

packages/hoc-input/src/formInput.js
```diff
@@ -73,13 +81,38 @@ const formInput = WrappedComponent => {
...
+        : inputProps;
+      const icon = composedProps.icon && (
+        <div className="icon">
+          <i className={composedProps.icon} />
+        </div>
+      );
+      const clearButton = isSearch &&
+        this.state.value && (
+          <Link className="clear" onClick={this.handleClearClick}>
+            <i className="wg-close-int" />
```
alcferreira: Close button icon should be `GREY_32` 
image

Set `cursor: pointer` to `i` tag.

packages/hoc-input/src/formInput.js
vNakamura: Actually, no.
Passing `this.inputRef` as target causes the browser to freeze. 👀

packages/hoc-input/src/formInput.js
vNakamura: It was `cursor:pointer` because it's inside a link. I've changed to `cursor:default` as that's how it works on the native search input. Should I use `pointer` even so?

alcferreira: Yeah. If you position the mouse on the left of the icon, it is a pointer. If you go over the icon, it s default.
We should keep consistency here.

Issue #230, opened 2018-01-29
Select box component


fczuardi:
design specs:
- https://app.zeplin.io/project/595a9cd3b401bf1876faab27/screen/59ee44fec5418ce9dc2815c0

also have 2 sizes, a small for the table filters and a bigger one for the table cells:

![screen shot 2018-01-29 at 10 48 12](https://user-images.githubusercontent.com/7760/35511128-f5b5ed3c-04e1-11e8-8b3a-43692bfa806f.png)
![screen shot 2018-01-29 at 10 48 05](https://user-images.githubusercontent.com/7760/35511129-f5d6d592-04e1-11e8-9785-b6d0629e29d9.png)

https://projects.invisionapp.com/d/main#/console/13080625/273581765/preview

alcferreira:
https://projects.invisionapp.com/d/main#/console/13207780/276092505/inspect

fczuardi:
More references:

https://invis.io/FHFPLVJQ4YM#/276092505_---_Selects
https://invis.io/NSFPLLZEM82#/276092492_Advanced_Search_Components

PR #235, opened 2018-02-01
Select component

alcferreira:
affects: @crave/farmblocks-input-select

ISSUES CLOSED: #230


fczuardi:
disabled text color is wrong:
screen shot 2018-02-06 at 15 40 43
screen shot 2018-02-06 at 15 35 36

https://projects.invisionapp.com/d/main#/console/13207780/276092505/preview

fczuardi:
autocomplete results with zero items are dropping shadow

screen shot 2018-02-06 at 15 42 31

alcferreira:
What should render when there is no item that matches the search criteria?

fgameiro:
It would be nice if the mouse cursor was 'text' to indicate the editable field.
select-edit

alcferreira:
@fczuardi Addressed

packages/input-select/README.md
```diff
@@ -0,0 +1,56 @@
...
+];
+
+class App extends Component {
+  render() {
+    return (
+      <Select
+        placeholder="Select fruit"
+        label="Fruit"
+        items={items}
+        onChange={action("onChange")}
```
fczuardi: copy/paste of this wont work because of action()

packages/input-select/README.md
```diff
@@ -0,0 +1,56 @@
...
+      />
+    );
+  }
+}
+
+render(<App />, document.getElementById('root'));
```
fczuardi: copy / paste of the usage on stackblitz didn't work for me

fczuardi:
is this the expected visual? 2 circles?

screen shot 2018-02-07 at 15 16 58

alcferreira:
The outer circle is from `Image` component. The inner is the image itself

packages/input-select/src/Select.js
```diff
@@ -0,0 +1,159 @@
...
+  }
+}
+
+Select.defaultProps = {
+  onChange: () => false
+};
+
+Select.propTypes = {
+  items: PropTypes.arrayOf(
+    PropTypes.shape({
```
fczuardi: I dont know about this fixed shape here, if the renderItem is used, items can have any shape

packages/input-select/src/Select.js
```diff
@@ -0,0 +1,159 @@
...
+  _renderInput(autoCompleteProps) {
+    const { ref, ...rest } = autoCompleteProps;
+
+    const { renderItem, ...inputProps } = this.props;
+
+    inputProps.validationMessages = this.state.isMenuOpen
+      ? []
+      : this.props.validationMessages;
+
+    return <EnhancedInput {...inputProps} {...rest} innerRef={ref} />;
```
fczuardi: we should have and option to render a select box where the input is not a text input IMHO, just like a regular html select box... but we can handle this on a separated enhancement issue

fgameiro:
Maybe this is a minor issue, mas when the search fails and the empty card appears, an error is thrown

captura de tela de 2018-02-07 15-43-33

packages/input-text/package.json
```diff
@@ -36,6 +30,6 @@
   },
   "dependencies": {
     "@crave/farmblocks-hoc-disabled-tooltip": "^1.0.1",
-    "@crave/farmblocks-hoc-input": "^1.1.1"
+    "@crave/farmblocks-hoc-input": "^1.2.0-alpha.4b468a5d"
```
fczuardi: lets change this back and trust that the semantic release will do the right thing :)

alcferreira:
@fgameiro nice catch. Addressed.

packages/hoc-input/src/formInput.js
```diff
@@ -1,5 +1,6 @@
 import * as React from "react";
 import PropTypes from "prop-types";
+import classNames from "classnames";
```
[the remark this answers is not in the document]
alcferreira: Yeah... I could concatenate it myself but I thought that it might need add more conditional classes in the future so decided to use classNames to make it a bit cleaner.
But no doubts that we need to break this hoc into smaller ones

packages/input-select/src/Select.js
alcferreira: Agreed

packages/input-select/README.md
alcferreira: Opsss

packages/input-select/src/Select.js
alcferreira: Sure... Did it this way to have a pattern when we use this component, if needed we can change this in the future for sure.

packages/input-text/package.json
alcferreira: I removed the commit

packages/input-select/src/Select.js
fczuardi: filed #243

Issue #252, opened 2018-02-15
Make theme customizable


vNakamura:
This is required for CraveFood/webapp/issues/2572
We need the ability to change the components' colors (mostly buttons) without having to rebuild the packages.

We'll probably use styled-component's `<ThemeProvider>`:
https://www.styled-components.com/docs/advanced#theming

fczuardi:
I like the idea of using styled-components ThemeProvider

PR #253, opened 2018-02-15
Enable custom themes for buttons

vNakamura:
affects: `@crave/farmblocks-button`

ISSUES CLOSED: #252


packages/button/src/styledComponents/Button.js
```diff
@@ -80,8 +80,8 @@ function paddingStyle(props) {
   `;
 }
 
-function neutralStyle(props) {
-  const { textColor, iconColor, textHoverColor, color } = colorTypes[NEUTRAL];
+function neutralStyle(props, theme) {
```
fczuardi: theme on a component child of ThemeProvider is already present in the props, I think we could keep just one argument here with a little tweak below on the place this function is called

packages/button/src/styledComponents/Button.js
```diff
@@ -105,10 +105,11 @@ function neutralStyle(props) {
 }
 
 function typeStyle(props) {
+  const theme = Object.assign({}, colorTypes, props.theme);
   if (props.type === NEUTRAL) {
-    return neutralStyle(props);
+    return neutralStyle(props, theme);
```
fczuardi: here (as discussed offline):
```
    return neutralStyle({...props, theme});
```

vNakamura:
@fczuardi  changes addressed

packages/button/src/styledComponents/Button.js
```diff
@@ -105,10 +105,11 @@ function neutralStyle(props) {
 }
 
 function typeStyle(props) {
+  const theme = Object.assign({}, colorTypes, props.theme);
```
alcferreira: Why not `spread operator` as you used on line 110? :P

`const theme = { ...colorTypes, ...props.theme };`

vNakamura: Because `props.theme` can be `undefined`

alcferreira: What if you set default `props.theme` to `{}`?

I personally prefer to use the spread operators, but if you guys prefer the `Object.assign` no problem

vNakamura: Yes! I prefer spread operators, too.
It didn't come to me that styled components could have defaultProps. I'll change it to spread.

PR #259, opened 2018-02-21
Hoc validation messages

### LINKED ISSUE #223: split hoc-input code
currently, the checkbox component are duplicating code that exists on the hoc-input, like the code to handle willReceivePorps and onChange. But we cant reuse the hoc because it makes too much stuff, it handle styling like shadows and borders that are not used by all form inputs, like the checkbox...

alcferreira:
closes #223


packages/hoc-validation-messages/src/withMessages.js
```diff
@@ -0,0 +1,119 @@
...
+
+export const withMessagesProps = {
+  errorIconSrc: PropTypes.string,
+  validationMessages: PropTypes.arrayOf(PropTypes.string),
+  onInvalid: PropTypes.func
+};
+
+const withMessages = WrappedComponent => {
+  return class ValidationMessages extends React.Component {
+    constructor(props) {
```
vNakamura: Any reason to have a `constructor` here?

You can set the state directly in the class:
```
state = { messages: this.props.validationMessages }
```

And you don't need to bind the methods if you use arrow functions. So instead of `onInvalid(event) {`, you can use `onInvalid = event => {` and remove the binding.
Or, if you need keep these as methods, you can bind outside the constructor:
`onInvalid = this.onInvalid.bind(this);`

Ref.: https://hackernoon.com/the-constructor-is-dead-long-live-the-constructor-c10871bea599

packages/hoc-validation-messages/README.md
```diff
@@ -0,0 +1,43 @@
...
+```
+
+## API
+
+Copy to clipboard | Property | Type | Default | Required | Description |
+|--------------------|------------------|-------------|----------|---------------------------------------------------------------------|
+| validationMessages | array of strings | [] | | the error messages to be
+rendered | | onInvalid | function | () => false | | | | errorIconSrc | string |
+| | the url for a custom icon to be displayed before all error messages | |
+invalid | boolean | false | | true if there are any validationMessages |
```
fgameiro: This table is not rendering correctly...

fgameiro: I think this `Copy to clipboard` is out of place

packages/hoc-validation-messages/src/withMessages.js
alcferreira: Yeah, no reason at all...

I'll change that, thanks

Issue #257, opened 2018-02-21
Removable tags component


alcferreira:
Develop removable tags component

![image](https://user-images.githubusercontent.com/17936244/36486300-f36dfd30-16fc-11e8-820a-d094e19e9dfc.png)

Specs: https://projects.invisionapp.com/d/main#/console/13207780/276092749/preview

PR #279, opened 2018-03-09
Removable tags component

fgameiro:
affects: @crave/farmblocks-removable-tags

ISSUES CLOSED: #257


packages/removable-tags/src/components/RemovableTags.js
```diff
@@ -0,0 +1,41 @@
...
+          <i className="wg-close" />
+        </div>
+      </StyledTag>
+    );
+  }
+}
+
+RemovableTags.propTypes = {
+  text: PropTypes.string.isRequired,
+  onClose: PropTypes.func
```
vNakamura: Minor: I'm not sure if `onClose` is the best name for this. Maybe `onRemove` would be more consistent with what the action and the component's name

packages/removable-tags/src/components/RemovableTags.js
```diff
@@ -0,0 +1,41 @@
...
+
+  onCloseHandler() {
+    this.setState({ isVisible: false });
+    this.props.onClose();
+  }
+
+  render() {
+    if (!this.state.isVisible) {
+      return null;
+    }
```
vNakamura: I don't know if this is necessary. The logic to hide the component should be handled by the container.

But let's check with the team what's their opinion before removing.

packages/removable-tags/src/components/RemovableTags.test.js
```diff
@@ -0,0 +1,17 @@
...
+import Adapter from "enzyme-adapter-react-16";
+
+import RemovableTags from "./RemovableTags";
+
+Enzyme.configure({ adapter: new Adapter() });
+
+describe("Removable tags", function() {
+  test("should call onClose when clicking on the icon", function() {
+    const onClickMock = jest.fn();
+    const wrapper = shallow(<RemovableTags onClose={onClickMock} />);
```
vNakamura: Minor: The function name is `onClickMock` but you are mocking a `onClose`. The naming should be consistent.

packages/removable-tags/src/components/styledComponent/StyledTag.js
```diff
@@ -0,0 +1,43 @@
...
+  color: white;
+  font-family: Lato;
+  font-weight: 600;
+
+  cursor: default;
+  &:hover {
+    border: 2px solid ${colorConstants.GREY_32};
+    padding: 5px 0px 7px 7px;
+    background-color: ${colorConstants.BLUE_CORN};
+  }
```
vNakamura: The interaction is a bit odd here. The tag gets a hover effect when the cursor is anywhere over the tag. But the only interactive part is the X:


The hover effect should happen only when the cursor is over the X:


You can achieve this by adding `pointer-events: none;` to the container and adding `pointer-events: auto;`  to the icon.

packages/removable-tags/src/components/RemovableTags.js
alcferreira: Agreed.

packages/removable-tags/src/components/RemovableTags.js
```diff
@@ -0,0 +1,41 @@
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+
+import StyledTag from "./styledComponent/StyledTag";
+
+class RemovableTags extends Component {
```
alcferreira: Rename package to `Tag`.

Make tag "removable" by receiving a property that says it is removable.
If not removable, do not render the close button.

packages/removable-tags/src/components/RemovableTags.js
alcferreira: Agreed.

packages/removable-tags/src/components/RemovableTags.js
```diff
@@ -0,0 +1,41 @@
...
+        <div className="icon" onClick={this.onCloseHandler}>
+          <i className="wg-close" />
+        </div>
+      </StyledTag>
+    );
+  }
+}
+
+RemovableTags.propTypes = {
+  text: PropTypes.string.isRequired,
```
alcferreira: Receive a property called `value` and return it when user clicks on the button to remove.
I am quite sure we will need this value on the FM setup milestone

packages/tag/src/components/Tag.story.js
```diff
@@ -3,35 +3,35 @@ import { storiesOf } from "@storybook/react";
...
-import RemovableTags from "./RemovableTags";
+import Tag from "./Tag";
 
 storiesOf("Removable-tags", "module")
   .add(
     "Default",
     withInfo()(() => (
       <div style={{ marginTop: "20px" }}>
-        <RemovableTags onRemove={action("tag removed")} text="THIS IS A TAG" />
+        <Tag onRemove={action("tag removed")} text="THIS IS A TAG" />
```
alcferreira: Refactor to `<Tag onRemove={action("onRemove")} text="THIS IS A TAG" />`
This way we can see which function was called on the action logger

packages/tag/src/components/Tag.story.js
```diff
@@ -0,0 +1,40 @@
+import React from "react";
+import { storiesOf } from "@storybook/react";
+import { withInfo } from "@storybook/addon-info";
+import { action } from "@storybook/addon-actions";
+
+import Tag from "./Tag";
+
+storiesOf("Removable-tags", "module")
```
alcferreira: "Tag"

packages/tag/src/components/Tag.story.js
```diff
@@ -0,0 +1,40 @@
...
+  .add(
+    "Default",
+    withInfo()(() => (
+      <div style={{ marginTop: "20px" }}>
+        <Tag onRemove={action("onRemove")} text="THIS IS A TAG" />
+      </div>
+    ))
+  )
+  .add(
+    "Not removable",
```
alcferreira: I think default tab is not removable... just pass a text and you have a tag.
So...
- [ ] rename this story to `Removable`
- [ ] remove `onRemove` property from `Default` story

packages/tag/src/components/Tag.test.js
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+import Enzyme, { shallow } from "enzyme";
+import Adapter from "enzyme-adapter-react-16";
+
+import Tag from "./Tag";
+
+Enzyme.configure({ adapter: new Adapter() });
+
+describe("Removable tags", function() {
```
alcferreira: `Tag`

alcferreira:
Please do not forget to use commitzen to commit.

image

packages/tag/src/components/Tag.story.js
```diff
@@ -0,0 +1,41 @@
...
+    ))
+  )
+  .add(
+    "Removable",
+    withInfo()(() => (
+      <div style={{ marginTop: "20px" }}>
+        <Tag text="You can remove this tag" onRemove={action("onRemove")} />
+      </div>
+    ))
+  )
```
vNakamura: Create a story with value.
We have a tag with value on the `multiple tags` story. But it would be better to have documentation for this prop.

PR #286, opened 2018-03-13
feat(tooltip): allow children to be rendered instead of only text

fgameiro:
affects: @crave/farmblocks-tooltip

ISSUES CLOSED: #285

### DIFF
--- a/packages/tooltip/src/Tooltip.story.js
+++ b/packages/tooltip/src/Tooltip.story.js
@@ -1,7 +1,7 @@
 import React from "react";
 
 import { storiesOf } from "@storybook/react";
-import Card from '@crave/farmblocks-card';
+import Card from "@crave/farmblocks-card";
 
 import Tooltip from "./Tooltip";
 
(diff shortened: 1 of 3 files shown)


packages/tooltip/src/styledComponents/Tooltip.js
```diff
@@ -13,7 +13,7 @@ const StyledTooltip = styled.div`
   box-shadow: 0 2px 2px 0 rgba(0, 0, 0, 0.16);
   border: solid 1px rgba(0, 0, 0, 0.16);
   border-radius: 4px;
-  white-space: pre;
```
vNakamura: Why was this removed?

I don't know if it's necessary, but let's check. @fczuardi  do you remember why was this rule added?

packages/tooltip/src/Tooltip.story.js
```diff
@@ -1,22 +1,37 @@
...
+        <Card>
+          this is a Card
+        </Card>
+      </Tooltip>
+    </div>
+  ))
+  .add("empty", () => (
+    <div style={{ position: "relative", width: "250px", border: "1px solid" }}>
+      <Tooltip isVisible={true} />
+    </div>
```
alcferreira: A `Tooltip` without content should not be rendered.

image

packages/tooltip/src/styledComponents/Tooltip.js
fgameiro: This was preventing a long text from line breaking, but I think I can leave it on if this effect is expected.

fgameiro: With `white-space: pre;`
captura de tela de 2018-03-14 09-11-08

Without:
captura de tela de 2018-03-14 09-10-44

vNakamura: We talked about this, but to keep documented:
Let's keep the `white-space: pre;`

PR #288, opened 2018-03-14
Center align option for tooltip

fgameiro:
affects:

ISSUES CLOSED: #287

### DIFF
--- a/packages/tooltip/src/styledComponents/Tooltip.js
+++ b/packages/tooltip/src/styledComponents/Tooltip.js
@@ -1,7 +1,21 @@
-import styled from "styled-components";
+import styled, { css } from "styled-components";
 import { colors } from "@crave/farmblocks-theme";
 
 const Container = styled.div`position: relative;`;
+
+const centerAlignment = css`
+  left: 50%;
+  transform: translateX(-50%);
+`;
+
+const alignX = coordinate => props => {
+  const { align } = props;
+  if (align === "center") {
+    return centerAlignment;
+  }
+
+  return css`${align}: ${coordinate}`;
+};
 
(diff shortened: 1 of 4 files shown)


packages/tooltip/src/styledComponents/Tooltip.js
```diff
@@ -1,8 +1,38 @@
...
+    ${align}: 0;
+    &:before {
+      ${align}: 7px;
+    }
+    &:after {
+      ${align}: 8px;
+    }
+  `;
+};
+
```
alcferreira: This works, but i think we can simplify this function in order to make it more readable.
I mean... on the `StyledTooltip` we have 3 places that needs align. So we need to go up and down on code to understand what it is aligning.

My proposal is to create a closure to store the x coordinate and align it accordinly:

```
const centerAlignment = css`
  left: 50%;
  transform: translateX(-50%);
`;

const alignX = coordinate => props => {
  const { align } = props;
  if (align === "center") {
    return centerAlignment;
  }

  return css`${align}: ${coordinate}`;
};
```

Then we can use it on the `StyledTooltip`: 

```
const StyledTooltip = styled.div`
  (...)

  ${alignX(0)};

  (...)

  &:before {
    border-bottom-color: rgba(0, 0, 0, 0.16);
    border-width: 8px;
    ${alignX("7px")};
  }

  &:after {
    border-bottom-color: #ffffff;
    border-width: 7px;
    ${alignX("8px")};
  }
`;
```

fgameiro: Beautiful. Thanks!

alcferreira:
Please fixup the commits (:

Issue #294, opened 2018-03-20
Improve Map Balloon animation performance


vNakamura:
Make changes so the transition will be processed by GPU

alcferreira:
See https://github.com/CraveFood/farmblocks/pull/295#issuecomment-376140499

PR #295, opened 2018-03-20
[WIP] Improve Balloon animation performance

vNakamura:
affects: @crave/farmblocks-map-balloon

Closes: #294


alcferreira:
Looks like it has doubled scripting time:

Before: 
23 to 67 fps 
image

After
13 fps
image

vNakamura:
Further reading about the `will-change` property, it's not suitable for this case as the animation runs at component mount. 😞

Issue #299, opened 2018-03-24
Improve Carousel performance


alcferreira:

alcferreira:
Before:
![image](https://user-images.githubusercontent.com/17936244/37866587-7a10178e-2f6b-11e8-9b92-1286f59acc3e.png)

Later:
![image](https://user-images.githubusercontent.com/17936244/37866592-8d059fda-2f6b-11e8-88b0-225a427bb056.png)

alcferreira:
After scaling instead of increase width and height:

![image](https://user-images.githubusercontent.com/17936244/37867318-1e250b6e-2f75-11e8-8613-96a0b8fb1566.png)

PR #300, opened 2018-03-24
Improve Carousel performance

alcferreira:
ISSUES CLOSED: #299

### DIFF
--- a/packages/carousel/src/Carousel.js
+++ b/packages/carousel/src/Carousel.js
@@ -72,7 +72,7 @@ class Carousel extends React.Component {
     return (
       <Container
         activeItem={this.state.activeItem}
-        itemConfig={{ ...itemConfig, ...defaultConfig }}
+        itemConfig={{ ...defaultConfig, ...itemConfig }}
         scale={this.props.scale}
       >
         <ul>
--- a/packages/carousel/src/Carousel.js
+++ b/packages/carousel/src/Carousel.js
@@ -9,7 +9,7 @@ const defaultConfig = {
   width: 656,
   height: 328,
   margin: 20,
-  fontSize: "88px",
+  fontSize: 88,
   displayTime: 4,
   transitionTime: 2,
   border: {
@@ -66,7 +66,7 @@ class Carousel extends React.Component {
                   width="100%"
                   height="100%"
                 />
-                <Text className="caption" title>
+                <Text size={itemConfig.fontSize} align="center" title>
                   {item.name}
                 </Text>
               </li>
@@ -88,7 +88,7 @@ class Carousel extends React.Component {
       width: PropTypes.number,
       height: PropTypes.number,
       margin: PropTypes.number,
-      fontSize: PropTypes.string,
+      fontSize: PropTypes.number,
       displayTime: PropTypes.number,
       transitionTime: PropTypes.number,
       border: PropTypes.shape({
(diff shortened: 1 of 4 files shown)


packages/carousel/src/styledComponents/Carousel.js
```diff
@@ -19,32 +31,28 @@ const Container = styled.div`
...
+    transition: transform ${props => props.itemConfig.transitionTime}s;
+    will-change: transform;
+
     li {
       margin: ${props => props.itemConfig.margin}px;
-      transition: all ${props => props.itemConfig.transitionTime}s;
       width: ${props => props.itemConfig.width}px;
       height: ${props => props.itemConfig.height}px;
 
+      transition: all ${props => props.itemConfig.transitionTime}s;
```
vNakamura: Maybe we'll get a little performance improvement if we set the transition only for the objects that we change (`transform` and `margin`)

packages/carousel/src/Carousel.js
```diff
@@ -27,31 +27,53 @@ class Carousel extends React.Component {
...
+    }
   };
 
   render() {
     const { imageSet, itemConfig } = this.props;
     return (
       <Container
         activeItem={this.state.activeItem}
-        itemConfig={{ ...defaultConfig, ...itemConfig }}
+        itemConfig={{ ...itemConfig, ...defaultConfig }}
```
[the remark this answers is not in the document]
alcferreira: Yeah! Now that you pointed out it, this change makes no sense... 
Default config should be the first argument... I'll drop this commit and dig into it to see what was going on

packages/carousel/src/styledComponents/Carousel.js
alcferreira: Good catch! It was changing opacity too. I'll change it

packages/carousel/src/Carousel.js
```diff
@@ -88,7 +88,7 @@ class Carousel extends React.Component {
       width: PropTypes.number,
       height: PropTypes.number,
       margin: PropTypes.number,
-      fontSize: PropTypes.string,
+      fontSize: PropTypes.number,
```
fczuardi: this is a breaking change, please inform this on the commit message

packages/carousel/src/Carousel.js
```diff
@@ -27,23 +27,44 @@ class Carousel extends React.Component {
...
     this.props.onChange(activeItem);
     this.setState({ activeItem });
   };
 
   componentDidMount = () => {
+    this.setInterval();
+  };
+
+  componentWillUnmount = () => {
+    this.clearInterval();
```
fczuardi: :+1:

packages/carousel/src/Carousel.js
```diff
@@ -65,7 +87,7 @@ class Carousel extends React.Component {
                   width="100%"
                   height="100%"
                 />
-                <Text className="caption" title>
+                <Text size={itemConfig.fontSize} align="center" title>
```
fczuardi: :+1:

packages/carousel/src/Carousel.js
alcferreira: I'll make a rebase on the commit

packages/carousel/src/Carousel.test.js
```diff
@@ -63,7 +63,7 @@ describe("Carousel", () => {
     expect(setIntervalSpy).toHaveBeenCalledTimes(1);
   });
 
-  test("clearInterval should be cleared on unmount", () => {
+  test("clearInterval should be called on unmount", () => {
```
fczuardi: :smile:

PR #306, opened 2018-03-26
Faster tests

### LINKED ISSUE #301: run the snapshots tests conditionally
The snapshots tests in our current test suite is the most expensive and time consuming of them. And we only need them to check for regressions at CI time and for completeness in our coverage reports. At development time they are mostly just a dead weight.

So this issue is for refactoring ou jest config files to only run the snapshot tests when a ``--coverage`` or a ``-u`` parameter is sent, freeing up the developer's machines resources on the regular ``test`` and ``test --watch`` daily activities.

fczuardi:
- closes #301


vNakamura:
This test is failing for me if I run all tests (`yarn test --watch` then press **a**)

image

Any idea why?

fczuardi:
The find is not returning the expected dom node? Maybe it is just a legit failing test... I am checking it.

vNakamura:
@fczuardi removing the `node_modules` folder and reinstalling all the dependencies solved the problem.

Issue #307, opened 2018-03-26
Add "auto" option for Tooltip align


vNakamura:
Usually, we can manually set the alignment of the tooltip as we know where it will appear. But now we have a case where it's position will change along the user interaction:

Sometimes it will be at the left of the screen:
![image](https://user-images.githubusercontent.com/1459283/37926993-2bbac458-310f-11e8-9d0e-520ba81b70f6.png)
Sometimes it will be ate the right:
![image](https://user-images.githubusercontent.com/1459283/37926959-165b12c0-310f-11e8-9d5d-7a3552d16065.png)

In order to keep the tooltip content visible, we need an option to automatically choose `left`  or `right` alignment based on the tooltip position.

PR #308, opened 2018-03-26
Tooltip auto align

vNakamura:
Closes: #307

### DIFF
--- a/packages/tooltip/src/Tooltip.js
+++ b/packages/tooltip/src/Tooltip.js
@@ -5,8 +5,7 @@ import values from "object.values";
 import { Container, StyledTooltip } from "./styledComponents/Tooltip";
 import alignments from "./constants/alignments";
 
-export const getAutoAlign = bodyWidth => containerRef => align => {
-  if (align !== alignments.AUTO) return align;
+export const getAutoAlign = bodyWidth => containerRef => {
   if (containerRef) {
     const { x } = containerRef.getBoundingClientRect();
     if (x > bodyWidth / 2) return alignments.RIGHT;
@@ -19,11 +18,11 @@ class Tooltip extends React.Component {
     align: this.props.align
   };
   componentDidMount = () => {
-    this.setState({
-      align: getAutoAlign(document.body.clientWidth)(this.containerRef)(
-        this.props.align
-      )
-    });
+    const align =
+      this.props.align === alignments.AUTO
+        ? getAutoAlign(document.body.clientWidth)(this.containerRef)
+        : this.props.align;
+    this.setState({ align });
   };
   render() {
     const content = this.props.children || this.props.text;
--- a/packages/tooltip/src/Tooltip.js
+++ b/packages/tooltip/src/Tooltip.js
@@ -5,14 +5,42 @@ import values from "object.values";
 import { Container, StyledTooltip } from "./styledComponents/Tooltip";
 import alignments from "./constants/alignments";
 
-const Tooltip = props => {
-  const content = props.children || props.text;
-  return (
-    <Container>
-      {content && <StyledTooltip {...props}>{content}</StyledTooltip>}
-    </Container>
-  );
+export const getAutoAlign = bodyWidth => containerRef => align => {
+  if (align !== alignments.AUTO) return align;
+  if (containerRef) {
+    const { x } = containerRef.getBoundingClientRect();
+    if (x > bodyWidth / 2) return alignments.RIGHT;
+  }
+  return alignments.LEFT;
 };
+
+class Tooltip extends React.Component {
+  state = {
+    align: this.props.align
+  };
+  componentDidMount = () => {
+    this.setState({
+      align: getAutoAlign(document.body.clientWidth)(this.containerRef)(
+        this.props.align
+      )
+    });
+  };
+  render() {
+    const content = this.props.children || this.props.text;
+
+    return (
+      <Container>
+        <div ref={element => (this.containerRef = element)}>
+          {content && (
+            <StyledTooltip {...this.props} align={this.state.align}>
+              {content}
+            </StyledTooltip>
+          )}
+        </div>
+      </Container>
+    );
+  }
+}
 
 Tooltip.propTypes = {
   text: PropTypes.string,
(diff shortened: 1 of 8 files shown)


packages/tooltip/src/Tooltip.js
```diff
@@ -1,28 +1,58 @@
...
+  return alignments.LEFT;
 };
 
+class Tooltip extends React.Component {
+  state = {
+    align: this.props.align
+  };
+  componentDidMount = () => {
+    this.setState({
+      align: getAutoAlign(document.body.clientWidth)(this.containerRef)(
```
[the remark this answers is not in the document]
vNakamura: hum... not sure I get what you're asking.

You think this should be customizable by a prop? If so, then we would have to find other way to get the element's position instead of `getBoundingClientRect`, which will be a little harder.

Or you ask this to understand why this is an argument if it will allways be the same? This is just to make it easier to test, so we wont need to mock `document.body`

packages/tooltip/src/index.js
```diff
@@ -1 +1,2 @@
 export { default } from "./Tooltip";
+export { default as alignments } from "./constants/alignments";
```
alcferreira: 👌

packages/tooltip/src/Tooltip.js
```diff
@@ -5,15 +5,43 @@ import values from "object.values";
...
 
-const Tooltip = props => {
-  const content = props.children || props.text;
-  return (
-    <Container>
-      {content && <StyledTooltip {...props}>{content}</StyledTooltip>}
-    </Container>
-  );
+export const getAutoAlign = bodyWidth => containerRef => align => {
+  if (align !== alignments.AUTO) return align;
```
alcferreira: I think it would be better to move this condition to `componentDidMount`

packages/tooltip/src/Tooltip.js
```diff
@@ -5,15 +5,43 @@ import values from "object.values";
...
 import alignments from "./constants/alignments";
 
-const Tooltip = props => {
-  const content = props.children || props.text;
-  return (
-    <Container>
-      {content && <StyledTooltip {...props}>{content}</StyledTooltip>}
-    </Container>
-  );
+export const getAutoAlign = bodyWidth => containerRef => align => {
```
alcferreira: On the `Tooltip` class we could have access to all these properties with no need to create this closure...
Maybe it would be simplified it it was on the class instance.
What you think? Any particular reasons for it to live here?

packages/tooltip/src/Tooltip.js
```diff
@@ -5,15 +5,43 @@ import values from "object.values";
...
+  if (align !== alignments.AUTO) return align;
+  if (containerRef) {
+    const { x } = containerRef.getBoundingClientRect();
+    if (x > bodyWidth / 2) return alignments.RIGHT;
```
alcferreira: Some comments would be nice to help us understand what's going on... (why `bodyWith / 2` and so on)

packages/tooltip/src/Tooltip.js
vNakamura: I made it like this for testability. So it's not necessary to inject the `containerRef` mock into the component nor manipulate the `document` global.

packages/tooltip/src/Tooltip.js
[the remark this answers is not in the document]
fczuardi: if we use document.body as the frame to calculate when the tooltip would go "outside of the limits" we are assuming that the limits are always the screen limits.

But we might have cases where we dont want to surpass the limits of a different container, a div with overflow hidden for example.

packages/tooltip/src/Tooltip.js
```diff
@@ -7,8 +7,9 @@ import alignments from "./constants/alignments";
 
 export const getAutoAlign = bodyWidth => containerRef => {
   if (containerRef) {
-    const { x } = containerRef.getBoundingClientRect();
-    if (x > bodyWidth / 2) return alignments.RIGHT;
+    const { x } = containerRef.getBoundingClientRect(); // Position of the element on viewport
+    const halfViewport = bodyWidth / 2;
+    if (x > halfViewport) return alignments.RIGHT;
```
alcferreira: Ok... So the logic here is: if the element is on the "second" half of the screen, it's gonna be right aligned no matter what:
image

I first thought that this feature was supposed to calculate if the tooltip content fits on the screen. If it doesn't fit the screen, then we would right align it.

If this is actually the expected behaviour, please let it documented.

packages/tooltip/src/Tooltip.js
alcferreira: Hmmm.... Still think we do not need a closure here since we always pass the body width and tests can be done on component did mount, but ok.

vNakamura:
@fczuardi @alcferreira for your both questions, https://github.com/CraveFood/farmblocks/pull/308#discussion_r177898660 and https://github.com/CraveFood/farmblocks/pull/308#discussion_r178030169:
This was supposed to be a quick adjustment to use in CraveFood/webapp/pull/2727 milestone.

I agree these changes will make the component more robust for other use cases. But it will take a time I don't have right now as I need to keep working on the milestone.
So I propose to postpone this PR and use the canary version in the fmkt setup. When the milestone is finished I'll get back to this and make the enhancements. Cool?

alcferreira:
All good, thanks

packages/tooltip/src/Tooltip.js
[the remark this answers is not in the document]
fczuardi: ok, no problems for me releasing this version as well, we can open an issue to add this improvement later

packages/tooltip/src/Tooltip.js
```diff
@@ -5,11 +5,18 @@ import values from "object.values";
...
+    const { right } = tooltipRef.getBoundingClientRect();
+
+    const bondariesNode =
+      bondariesSelector && tooltipRef.closest(bondariesSelector);
+
+    const maxRight =
+      (bondariesNode && bondariesNode.getBoundingClientRect().right) ||
+      window.innerWidth;
+
+    if (right >= maxRight) return alignments.RIGHT;
```
vNakamura: @alcferreira @fczuardi what you guys think of this approach?

We're getting the right border of the tooltip and comparing it to the given container right border.
If no `bondariesSelector` is given, the limit will be the viewport.

alcferreira: Really nice! I like it better now

PR #317, opened 2018-04-02
fix(Carousel): custom fontsize and border

### LINKED ISSUE #316: Fix custom fontsize and custom border on Carousel

alcferreira:
closes #316


fczuardi:
looks ok, but still dont cover a custom config that overrides just one of the nested attributes, like ``border.color`` for example.

I  wonder if it would be worth adding a small dependency like xtend here...
https://github.com/Raynos/xtend

But that is something we can address as enhancement later too

Issue #318, opened 2018-04-04
(Link) Line break with icon when href property is not provided


fgameiro:
Currently we are rendering the children as `div `, on Link Component, when no `href` is provided. This is causing a line break when using with `leftIcon`:

![captura de tela de 2018-04-04 11-47-49](https://user-images.githubusercontent.com/17071199/38315083-14eb5d14-37fe-11e8-80d5-c640b7d16e1f.png)

PR #319, opened 2018-04-04
Replace div for span when no href is provided on Link

fgameiro:
Closes: #318

### DIFF
--- a/packages/link/src/components/Link.js
+++ b/packages/link/src/components/Link.js
@@ -35,7 +35,7 @@ const Link = props => {
     displayBlock
   };
 
-  const componentType = linkProps.href ? "a" : "div";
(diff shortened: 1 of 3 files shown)


packages/link/src/components/Link.js
```diff
@@ -35,7 +35,7 @@ const Link = props => {
     displayBlock
   };
 
-  const componentType = linkProps.href ? "a" : "div";
+  const componentType = linkProps.href ? "a" : "span";
```
fczuardi: this is a breaking change, please change the commit message so semantic version can bump the major version of the package when published

fczuardi: ok, @vNakamura explained on Slack that the broken behaviuor was introduced earlier and this is actually the fix

PR #329, opened 2018-04-12
Add a right icon option to farmblocks-link

### LINKED ISSUE #326: Right icon option for Farmblocks-Link
We should be able to add a right icon to a Link, in addition to the existing left icon option.

fgameiro:
closes #326

### DIFF
--- a/packages/link/src/components/Link.story.js
+++ b/packages/link/src/components/Link.story.js
@@ -38,6 +38,14 @@ storiesOf("Link/Featured", module)
     ))
   )
   .add(
+    "with right icon",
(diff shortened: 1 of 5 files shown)


packages/link/src/components/Link.story.js
```diff
@@ -37,6 +37,14 @@ storiesOf("Link/Featured", module)
       </Link>
     ))
   )
+  .add(
+    "with right icon",
```
vNakamura: Add stories like this in `Link.Neutral.story.js` and `Link.Normal.story.js`

Issue #342, opened 2018-05-08
Add innerRef to Dropdown


alcferreira:
And pass it to Button

PR #343, opened 2018-05-08
feat(Dropdown): add innerRef property

alcferreira:
affects: @crave/farmblocks-dropdown

ISSUES CLOSED: #342

### DIFF
--- a/packages/dropdown/src/components/Dropdown.story.js
+++ b/packages/dropdown/src/components/Dropdown.story.js
@@ -107,4 +107,18 @@ storiesOf("Dropdown/Component", "module")
         <DropdownItem value={3} text="Strawberry" />
       </Dropdown>
     ))
+  )
+  .add(
+    "With ref",
+    withInfo()(() => (
+      <Dropdown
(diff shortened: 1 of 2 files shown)


packages/dropdown/src/components/Dropdown.story.js
```diff
@@ -6,6 +6,22 @@ import { action } from "@storybook/addon-actions";
 import Dropdown from "./Dropdown";
 import DropdownItem from "./DropdownItem";
 
+class WithRef extends React.Component {
+  render() {
+    return (
+      <Dropdown
+        text="Select fruit"
+        handleSelection={action("handleSelection")}
+        innerRef={node => (this.triggerButton = node)}
```
vNakamura: How about we use this instead?
```javascript
innerRef={action("innerRef")}
```

It's not exactly the intended use for `@storybook/addon-actions` but at least we a feedback that the ref is working.

alcferreira: Nice ideia. Will do that, thanks!

Issue #346, opened 2018-05-09
Render images on Select Input


alcferreira:
https://projects.invisionapp.com/d/main#/console/13207780/276092503/preview

PR #347, opened 2018-05-09
Select Input w/ image

vNakamura:
- Closes: #346

### DIFF
--- a/packages/input-select/src/Select.js
+++ b/packages/input-select/src/Select.js
@@ -16,6 +16,8 @@ import withMessages, {
 } from "@crave/farmblocks-hoc-validation-messages";
 
 import EmptyCard from "./components/EmptyCard";
+import ItemImage from "./styledComponents/ItemImage";
+import ItemContainer from "./styledComponents/ItemContainer";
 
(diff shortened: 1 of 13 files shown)


packages/input-select/src/Select.js
```diff
@@ -174,29 +154,36 @@ class Select extends React.Component {
...
+  static defaultProps = {
+    onChange: () => false
+  };
+
+  static propTypes = {
+    items: PropTypes.arrayOf(
+      PropTypes.shape({
+        value: PropTypes.string,
+        label: PropTypes.string,
+        image: PropTypes.string
```
alcferreira: Update docs with the new image prop

packages/input-select/src/Select.js
```diff
@@ -174,29 +154,36 @@ class Select extends React.Component {
...
+      })
+    ).isRequired,
+    value: PropTypes.string,
+    onChange: PropTypes.func,
+    renderItem: PropTypes.func,
+    noResultsMessage: PropTypes.string,
+    disableSearch: PropTypes.bool,
```
alcferreira: `disableSearch` is not on the docs either. Please add it too

Issue #243, opened 2018-02-07
Support simple select boxes (without text input)


fczuardi:
Sometimes a select box input is just a way to choose between few options, like two or three available choices, and having a field that you can type to filter out options dont make much sense.

This issue is about having a way to us the select component like a regular html select where all options are shown and no typing is allowed.

PR #351, opened 2018-05-10
Add read only prop to hoc-input

fgameiro:
closes: #243

### DIFF
--- a/packages/input-select/src/Select.js
+++ b/packages/input-select/src/Select.js
@@ -93,13 +93,20 @@ class Select extends React.Component {
   _renderInput(autoCompleteProps) {
     const { ref, ...rest } = autoCompleteProps;
 
-    const { renderItem, ...inputProps } = this.props;
+    const { renderItem, disableSearch, ...inputProps } = this.props;
 
     inputProps.validationMessages = this.state.isMenuOpen
       ? []
       : this.props.validationMessages;
 
-    return <EnhancedInput {...inputProps} {...rest} innerRef={ref} />;
+    return (
+      <EnhancedInput
+        readOnly={disableSearch}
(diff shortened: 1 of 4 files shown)


packages/input-select/src/Select.story.js
```diff
@@ -95,4 +95,18 @@ storiesOf("Select Input", "module")
...
+  .add(
+    "Read Only",
+    withInfo()(() => (
+      <Wrapper>
+        <Select
+          placeholder="Select fruit"
+          label="Fruit"
+          items={items}
+          onChange={action("onChange")}
+          readOnly
```
alcferreira: `readOnly` gives me the idea that the Select is disabled. The goal of this property is to disable the search feature.

Our suggestion is to receive a property called `disableAutoComplete` or something like that and disable the formInput based on this property.

alcferreira: Or even `disableSearch`

packages/input-select/src/Select.js
```diff
@@ -157,6 +164,7 @@ Select.propTypes = {
   onChange: PropTypes.func,
   renderItem: PropTypes.func,
   noResultsMessage: PropTypes.string,
+  disableAutocomplete: PropTypes.bool,
```
fczuardi: I dont know If I like this name, this property disables typing in the input

fgameiro:
What do you guys think of `disableTyping`? `disableEdit`?
I actually like both better... that propert prevents the typing, which consequently avoids the autocompletion.

alcferreira:
I like `disableSearch` better, but it's up to you guys

fgameiro:
`disabledSearch` then :)

Issue #341, opened 2018-05-07
Amount selectors 


alcferreira:
https://projects.invisionapp.com/d/main#/console/13207780/276092508/preview

fczuardi:
The update of this font on our storybook is tracked by https://github.com/CraveFood/farmblocks/issues/369

fczuardi:
In the design specs, the increment and decrement buttons are of size 48x48 px, the existing button component dont support such size, issue #370 tracks this enhancement, but we should check with @vitorpinho376 to see if this new large size is correct or if we should be using the MEDIUM one (34px)

PR #368, opened 2018-05-17
[WIP]Amount selectors component

### LINKED ISSUE #369: update fonticon used by storybook
We need to update the font used by storybook and the global css to include the new icon introduced with https://github.com/CraveFood/webapp/issues/2814 (minus icon)

fgameiro:
affects: @crave/farmblocks-amount-selectors

ISSUES CLOSED: #341 
- Closes #369


packages/amount-selectors/src/components/AmountSelectors.js
```diff
@@ -0,0 +1,125 @@
...
+    );
+  }
+
+  static propTypes = {
+    value: PropTypes.number,
+    step: PropTypes.number,
+    min: PropTypes.number,
+    max: PropTypes.number,
+    onChange: PropTypes.func,
+    disableEdit: PropTypes.bool
```
alcferreira: looks like user will never be able to edit the value with this flag on.
Can we get another name?

packages/amount-selectors/src/components/AmountSelectors.js
```diff
@@ -0,0 +1,125 @@
...
+      this.setState({
+        value: parsedValue,
+        validationMessages: []
+      });
+
+      return this.props.onChange(parsedValue);
+    }
+
+    this.setState({
+      validationMessages: ["Invalid value"]
```
alcferreira: > Type something
> Clear everything


image

packages/amount-selectors/src/components/AmountSelectors.js
```diff
@@ -0,0 +1,125 @@
...
+          noTooltip
+        />
+        <div className="inputContainer">
+          <InputText
+            value={this.state.value}
+            readOnly={this.props.disableEdit}
+            onKeyDown={this.onKeyDown}
+            onChange={this.onChange}
+            size={4}
+            validationMessages={this.state.validationMessages}
```
alcferreira: We can have validation messages coming from props as well.

packages/amount-selectors/src/components/AmountSelectors.test.js
```diff
@@ -0,0 +1,120 @@
...
+
+describe("Amount selectors", () => {
+  describe("Buttons", () => {
+    test("decrement button should decrease the state value by 1", () => {
+      const initialValue = 2;
+      const expectedValue = 1;
+      const component = shallow(<AmountSelectors value={initialValue} />);
+      component
+        .children("Button")
+        .at(0)
```
alcferreira: Can't we use a more descriptive selector? Getting the first position of an arrays seems weird. It works, but what if the children order changes at some point?

packages/amount-selectors/src/components/AmountSelectors.js
fgameiro: Sure... disableTyping?

alcferreira:
We have a small version as well
https://projects.invisionapp.com/d/main#/console/13207780/276092510/inspect

packages/amount-selectors/src/components/AmountSelectors.js
```diff
@@ -0,0 +1,120 @@
...
+    this.onChange(Math.min(value, this.props.max), this.updateDisplayValue);
+  };
+
+  render() {
+    return (
+      <Wrapper size={this.props.size}>
+        <Button
+          className="decreaseButton"
+          type={buttonTypes.SECONDARY}
+          size={this.props.size}
```
vNakamura: 🤔 The `Button` component receives a `selectorSizes` value? Don't buttons have their own constants `buttonSizes`?
They may be identical now but that could not be true in the future.

How about we create a helper function:
```javascript
const mapSelectorSizeToButtonSize = size => ({
  [selectorSizes.SMALL]: buttonSizes.SMALL,
  [selectorSizes.MEDIUM]: buttonSizes.MEDIUM
})[size];
```

packages/amount-selectors/src/components/AmountSelectors.js
```diff
@@ -0,0 +1,120 @@
...
+            max={this.props.max}
+            step={this.props.step}
+            value={this.state.displayValue}
+            readOnly={this.props.disableTyping}
+            onKeyDown={this.onKeyDown}
+            onChange={this.onChange}
+            onBlur={this.updateDisplayValue}
+            size={4}
+            validationMessages={this.state.validationMessages}
+            fontSize={fontSizes[this.props.size]}
```
vNakamura: The same comment from the `buttonSizes` applies here.

packages/amount-selectors/src/components/AmountSelectors.js
alcferreira: Actually we can remove the `validationMessages`

packages/amount-selectors/src/components/AmountSelectors.js
```diff
@@ -62,10 +72,11 @@ class AmountSelectors extends React.Component {
             min={this.props.min}
             max={this.props.max}
             step={this.props.step}
-            value={this.state.value}
+            value={this.state.displayValue}
```
vNakamura: Couldn't we simply use `this.state.value.toFixed(2)` here instead of having this `displayValue` state?
Am I missing something?

PR #380, opened 2018-05-30
Fix amount invalid style 

### LINKED ISSUE #379: Replace native css style for input:invalid on Amount Selectors (Firefox)
Another issue that happens on firefox 60 / Linux:

image

**Steps to reproduce**
- Open the storybook story for amount selector with step size 0.5
- Type a number that isnt multiple of that step (ex: 2.1)
- Press Tab to move the focus elsewhere

**Expected results**
Either:
- The textfield should use farmblocks textfield red border style, the same one for when validationMessages is not empty.
- Or just hide the browser red box shadow

fczuardi:
- closes #379

### DIFF
--- a/packages/amount-selectors/README.md
+++ b/packages/amount-selectors/README.md
@@ -18,6 +18,7 @@ npm install @crave/farmblocks-amount-selectors
 | max | maximum amount available | number |
 | onChange | function to handle the input onChange event | function |
 | disableTyping | disables direct typing | bool |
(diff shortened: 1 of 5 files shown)


alcferreira:
On firefox
image

fczuardi:
the problem above was fixed in #378 and is gone after the last rebase

alcferreira:
Could you please verify what is the proper behaviour when the step is invalid?
IMO we should not disable the buttons (current behaviour on webapp).

PR #382, opened 2018-05-31
Fix amount prop update

### LINKED ISSUE #381: Amount selector, value should change if the props value is update after mount

fczuardi:
- closes #381

### DIFF
--- a/packages/amount-selectors/src/components/AmountSelectors.js
+++ b/packages/amount-selectors/src/components/AmountSelectors.js
@@ -49,13 +49,13 @@ class AmountSelectors extends React.Component {
     return this.props.onChange(validValue);
   };
 
-  updateDisplayValue = () => {
-    const parsedValue = parseFloat(this.state.value) || 0;
-    const validValue = Math.min(
-      this.props.max,
-      Math.max(this.props.min, parsedValue)
-    );
+  validValue = value => {
+    const parsedValue = parseFloat(value) || 0;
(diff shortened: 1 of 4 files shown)


packages/amount-selectors/src/components/AmountSelectors.js
```diff
@@ -57,6 +57,10 @@ class AmountSelectors extends React.Component {
     this.onChange(Math.min(value, this.props.max), this.updateDisplayValue);
   };
 
+  componentWillReceiveProps(nextProps) {
+    this.onChange(nextProps.value, this.updateDisplayValue);
```
vNakamura: Can't this cause an infinite loop if `props.onChange` updates `props.value`?

Also, `componentWillReceiveProps` will be deprecated in React 17. Let's avoid using it for new code:
https://reactjs.org/blog/2018/03/29/react-v-16-3.html#component-lifecycle-changes
https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops

alcferreira: @fczuardi

fczuardi: regarding the possible loop, a listener for a value change that also immediately sets the value feels wrong, but anyways, the existing component dont trigger an onChange, so I guess it is ok if we dont trigger one as well, I will just update state without calling the change handlers.

PR #388, opened 2018-06-07
Fix search field style

### LINKED ISSUE #387: Fix Search Field selected style & onchange/selection flow
https://projects.invisionapp.com/d/main#/console/13207780/276092494/preview

Current:
image

Expected:
image

vNakamura:
- Closes  #387


alcferreira:
Just caught a bug/weird behavior: 

`SearchField` has 2 events that change its values: `onChange` and `onSelect`. These events should be consistent with the flow:

Example:
- `onChange` and `onSelect` should be dispatched with an empty value when user starts typing and then leaves the field without selecting any items:

- `onChange` and `onSelect` should have the same behaviour as the above when user clicks on the edit icon:

PR #393, opened 2018-06-13
[WIP] Table rowgroups

### LINKED ISSUE #391: Make table row expandable/collapsible
image

image

fczuardi:
- closes #391


fczuardi:
still missing for the patch is the small blue "L" icon + padding on the rows that are grouped

packages/table/README.md
```diff
@@ -58,6 +58,8 @@ core of the whole thing, it should be an array of objects, each item representin
 | selectionHeader | function ``(selectedRows, clearFunction) => React.node`` | a function that receives an array of selected rows data plus a function to clear selection; and should return a react node to be rendered as an action bar on top of the table, there is a helper component SelectionBar that can be used as the return of this function, or you can create your own... see the stories files for usage examples |
 | width | string | use this to manually change the width of the table |
 | rowHeight | string | use this to manually change the height of the body rows of the table, the package exports a set of named values as ``rowHeights``, you can import them and use ``rowHeights.SMALL`` to have a more compact table |
+| rowGroupKey | string | if you have rows that contains sub-rows as a list under a key, you can pass this property with the name of the key, to have a table with row groups generated |
+| collapsed | boolean | if row groups are used, this flag will add a button column with buttons that works as expand/collapse toggle on the start of row groups. The groups will start collapsed. |
```
vNakamura: This could be splited in 2 props: `collapsible`  and `startCollapsed`. What do you think?

packages/table/src/Table.js
```diff
@@ -12,10 +12,34 @@ import { rowHeights } from "./constants";
...
 
+  componentDidMount() {
+    const { rowGroupKey } = this.props;
+    const hasSubRows = row =>
+      rowGroupKey && row[rowGroupKey] && row[rowGroupKey].length;
+
+    // iterate over all rows and sub-rows to create
+    // a hash table of rows indexed by keys in the
+    // format "<index>,<subIndex>"
+    const rowsMap = this.props.data.reduce((entries, row, index) => {
```
alcferreira: What happens when rows are added to `props.data` after the component is mounted?

packages/table/src/Table.js
```diff
@@ -59,30 +87,94 @@ class Table extends React.Component {
...
+          );
+        })}
       </StyledTable>
     );
   }
 
+  _renderRow = (row, index, subIndex = "", group = false) => {
+    const { selectableRows, collapsed, children } = this.props;
+    const rowKey = `${index},${subIndex}`;
+    const selected = this.state.selectedRows.indexOf(rowKey) !== -1;
```
vNakamura: I prefer`includes` over `indexOf` because it's declarative.

A downside is that it's not supported by IE.
We could add an indication of this restriction in the readme, so the use of a polyfill would be decided at the application using this component.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

Anyway, this is minor and I'm OK if we decide to keep `indexOf`.

packages/table/src/Table.js
```diff
@@ -59,30 +87,94 @@ class Table extends React.Component {
...
+
+  expandToggleClicked = index => {
+    const oldExpandedRows = this.state.expandedRows;
+    const expandedIndexOf = oldExpandedRows.indexOf(index);
+    const expandedRows =
+      expandedIndexOf !== -1
+        ? [
+            ...oldExpandedRows.slice(0, expandedIndexOf),
+            ...oldExpandedRows.slice(expandedIndexOf + 1)
+          ]
```
alcferreira: image

We could use filter here...
```
const expandedRows = 
  expandedIndexOf !== -1
    ? oldExpandedRows.filter((element, index) => index !== expandedIndexOf)
    : [...oldExpandedRows, index]
```

packages/table/src/Table.js
fczuardi: would break selection checkboxes :P

fczuardi: actually, I dont know, it would probably continue working ok, need to test it

fczuardi: actually is the selectionData passed to selectionHeader that breaks, it wont pass a selected row if the data about that row was added afterwards, I will update the index on component update to fix this (and write tests)

packages/table/README.md
fczuardi: how will this be used? will we have tables that starts expanded? if so, maybe we can have an `expanded` for that case

packages/theme/src/colors.js
```diff
@@ -2,8 +2,10 @@ module.exports = {
   CARBON: "#2f313a",
   OYSTER: "#59636f",
   SUGAR: "#f6f8f9",
+  DEMERARA_SUGAR: "#e9ebec",
   RED_ORANGE: "#ff4411",
   INDIGO_MILK_CAP: "#3498db",
+  INDIGO_MILK_CAP_06: "rgba(52,152,219,0.06)",
```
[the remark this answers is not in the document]
fczuardi: fixed on d96841023e21710e1c791862763429a6b0338516

packages/table/src/Table.js
fczuardi: fixed on 01af33dcc8df661329844193494f741e2e6dc4dc

packages/table/README.md
vNakamura: Yeah, we don't need to worry about it now

Issue #394, opened 2018-06-13
Search Fields - don't show menu if items is an empty array


vNakamura:
If `items` prop is set to `[]`, the menu will appear when there's focus on the input.
It has 0px height but it's still visible because of the border and shadow.

![image](https://user-images.githubusercontent.com/1459283/41376534-e0e12b3e-6f2f-11e8-9d5b-c739adf6a6b5.png)

PR #395, opened 2018-06-14
Fix SearchField with empty items array

vNakamura:
- Closes: #394

### DIFF
--- a/packages/search-field/src/components/Menu.js
+++ b/packages/search-field/src/components/Menu.js
@@ -53,11 +53,9 @@ const Menu = ({
         onReachEnd={onScrollReachEnd}
         ref={innerRef}
       >
-        {items.length
-          ? items.map((item, index) =>
-              renderItem(item, onItemClick, index === highlightedIndex)
-            )
-          : null}
+        {items.map((item, index) =>
+          renderItem(item, onItemClick, index === highlightedIndex)
+        )}
         {footer}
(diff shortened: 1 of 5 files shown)


packages/search-field/src/components/Menu.js
```diff
@@ -44,21 +44,28 @@ const Menu = ({
...
+}) => {
+  if (!footer && !items.length) return null;
+  return (
+    <DropdownMenuWrapper onMouseDown={onMenuMouseDown}>
+      <ScrollBox
+        maxHeight={maxMenuHeight}
+        onReachEnd={onScrollReachEnd}
+        ref={innerRef}
+      >
+        {items.length
```
alcferreira: I guess it's safe to remove this condition, right? (:

packages/search-field/src/components/Menu.js
```diff
@@ -44,21 +44,28 @@ const Menu = ({
...
-      {items &&
-        items.map((item, index) => {
-          return renderItem(item, onItemClick, index === highlightedIndex);
-        })}
-      {footer}
-    </ScrollBox>
-  </DropdownMenuWrapper>
-);
+}) => {
+  if (!footer && !items.length) return null;
```
fczuardi: I am not super found of bracket-less ifs, but that's a personal preference and prettier is ok with that, so no problem

Issue #392, opened 2018-06-12
Popover component


alcferreira:
Develop popover component

- [x] We should receive a trigger component (in this case a `Link`) and open the popover on click.
- [x] One click outside the container should dismiss the popover.
- [x] Popover should have a custom event to be dismissed.
- [x] Popover's contents should be whatever user wants to render.
- [x] Popover can be aligned to left, center and right.

![image](https://user-images.githubusercontent.com/17936244/41298753-53be5c38-6e38-11e8-9338-c0185f2bf5e0.png)

![image](https://user-images.githubusercontent.com/17936244/41298502-e3c244bc-6e37-11e8-9b17-15ac8a26fd9d.png)

PR #396, opened 2018-06-14
Popover

alcferreira:
affects: @crave/farmblocks-popover

ISSUES CLOSED: #392

### DIFF
--- a/packages/popover/CHANGELOG.md
+++ b/packages/popover/CHANGELOG.md
@@ -1,0 +1,13 @@
+# Change Log
+
+All notable changes to this project will be documented in this file.
+See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
+
+<a name="0.2.0-alpha.c35333c6"></a>
+# 0.2.0-alpha.c35333c6 (2018-06-15)
+
+
+### Features
+
+* **popover:** initial scaffold ([03e43dc](https://github.com/CraveFood/farmblocks/commit/03e43dc)), closes [#392](https://github.com/CraveFood/farmblocks/issues/392)
+* **Popover:** fully functional component ([eb244a6](https://github.com/CraveFood/farmblocks/commit/eb244a6))
--- a/packages/popover/src/Popover.js
+++ b/packages/popover/src/Popover.js
@@ -1,7 +1,65 @@
 import * as React from "react";
+import PropTypes from "prop-types";
+import styled from "styled-components";
+import values from "object.values";
+import Tooltip, { alignments } from "@crave/farmblocks-tooltip";
 
-const Popover = props => {
-  return <div />;
-};
+const Container = styled.div`display: inline-block;`;
+
+class Popover extends React.Component {
+  state = {
+    isVisible: false
+  };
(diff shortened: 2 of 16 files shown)


packages/popover/src/Popover.story.js
```diff
@@ -0,0 +1,70 @@
...
+      />
+    ))
+  )
+  .add(
+    "Custom padding",
+    withInfo()(() => (
+      <Popover
+        trigger={<button>Trigger</button>}
+        content={dismiss => (
+          <div>
```
fczuardi: I think you missed the custom padding :)

alcferreira: oops

packages/tooltip/src/Tooltip.js
```diff
@@ -58,13 +58,16 @@ Tooltip.propTypes = {
...
+  showArrow: PropTypes.bool,
+  padding: PropTypes.string
 };
 
 Tooltip.defaultProps = {
   isVisible: true,
   align: alignments.LEFT,
-  zIndex: 1000
+  zIndex: 1000,
+  showArrow: true
```
fczuardi: if the default is to display the arrow, maybe the property could be a flag to hide it instead, `hideArrow` that if present hides it.

packages/popover/CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
...
+See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
+
+<a name="0.2.0-alpha.c35333c6"></a>
+# 0.2.0-alpha.c35333c6 (2018-06-15)
+
+
+### Features
+
+* **popover:** initial scaffold ([03e43dc](https://github.com/CraveFood/farmblocks/commit/03e43dc)), closes [#392](https://github.com/CraveFood/farmblocks/issues/392)
+* **Popover:** fully functional component ([eb244a6](https://github.com/CraveFood/farmblocks/commit/eb244a6))
```
fczuardi: I am assuming that this got generated because you had to manually publish the package, right?

alcferreira: yes... I shouldn't have commited but...

packages/popover/src/Popover.js
```diff
@@ -0,0 +1,65 @@
...
+  handleOuterClick = event => {
+    if (this.popover.contains(event.target)) {
+      return;
+    }
+
+    this.hide();
+  };
+
+  toggle = () => {
+    this.setState((prevState, props) => ({ isVisible: !prevState.isVisible }));
```
fgameiro: This `props` argument can be removed?

packages/tooltip/README.md
```diff
@@ -1,25 +1,38 @@
-# Farmblocks Tooltip
+# Farmblocks popover
```
vNakamura: It seems like you got the `README.md` files from the `tooltip`  and `popover` packages mixed

alcferreira: Fixed, thanks!

Issue #414, opened 2018-06-22
SearchField should be focused after click on edit/clear icon


alcferreira:
SearchField should be focused after click on edit/clear icon

![2018-06-22 12 37 40](https://user-images.githubusercontent.com/17936244/41785469-1bb83942-7619-11e8-8ecc-54b0f6d25fae.gif)

PR #418, opened 2018-06-25
[WIP](Search Field): fix focus issues when editing or removing value

fgameiro:
affects: @crave/farmblocks-search-field

## Closes Issues

ISSUES CLOSED: #414

### Reviewer 1

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### Reviewer 2

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs


packages/hoc-input/src/formInput.js
```diff
@@ -90,9 +90,19 @@ const formInput = WrappedComponent => {
...
+      // clearButton. In this case, SearchField would lose focus because of
+      // its onBlur function, which is fired before this clearButton's onClick,
+      // and hence handleClearClick would not be called.
       const clearButton = (clearable || isSearch) &&
         this.state.value && (
-          <Link className="clear" onClick={this.handleClearClick}>
+          <Link
+            className="clear"
+            onClick={this.handleClearClick}
+            onMouseDown={this.preventBlur}
```
fczuardi: maybe we could use only the onMouseDown and put the preventDefault inside handleClearClick

packages/search-field/src/SearchField.js
```diff
@@ -37,9 +37,17 @@ class SearchField extends React.Component {
...
   );
+
   onSearchChange = event => {
     const { value } = event.target;
 
+    // We are forcing an update to render an input,
+    // otherwise we'd have a div instead,
+    // and calling focus on it
+    if (value === "") {
+      this.forceUpdate(() => this.input.focus());
```
fczuardi: This is difficult to understand, even with the comment. The focus is necessary after a click on the pencil, that reason should be mentioned, or alternatively the focus should be an option to pass to the input element upon mounting, I have pasted an example on slack.

fgameiro: Talking to @vNakamura we liked your suggestion as it's clearer about what's happening. Please see the next commit.

We also found and fixed another related bug, which was happening when typing something and then clicking anywhere outside the typing area: the component blinks (caused by input hoc) and the text is cleared. See gif.

search-focus-v3

packages/search-field/src/SearchField.js
```diff
@@ -212,8 +213,17 @@ class SearchField extends React.Component {
     const { focused, selectedItem } = this.state;
 
     const Input = selectedItem ? ReadOnly : EnhancedInput;
+
     Input.displayName = "Input";
 
+    // Focus the input component after
+    // selecting an item and then clearing it
+    // by clicking on the pencil icon
+    const autoFocus =
```
vNakamura: I think this comment is still a bit confusing. My suggestion:
```javascript
// When there's an item selected and we click the edit icon,
// the input ref points to the StaticInput instance, which has no focus().
// By setting the component focused prop, the regular
// input will get focus when it gets mounted.
```

fgameiro: Thanks!

packages/search-field/src/SearchField.js
```diff
@@ -212,8 +213,18 @@ class SearchField extends React.Component {
...
+
     Input.displayName = "Input";
 
+    // When there's an item selected and we click the edit icon,
+    // the input ref points to the StaticInput instance, which has no focus().
+    // By setting the component focused prop, the regular
+    // input will get focus when it gets mounted.
+    const autoFocus =
+      this.input !== undefined &&
+      this.input.nodeName !== "INPUT" &&
```
alcferreira: I believe this condition should be:
```
this.input && this.input.nodeName !== 'INPUT' && !selectedItem
```

An error could be thrown if `this.input = null` on line 225

packages/search-field/src/SearchField.js
```diff
@@ -224,6 +235,7 @@ class SearchField extends React.Component {
           onChange={this.onSearchChange}
           type={selectedItem ? "text" : "search"}
           clearable
+          focused={autoFocus}
```
alcferreira: Is this covered by storyshots? If not, we must add an unit test here...

packages/hoc-input/src/formInput.js
```diff
@@ -110,6 +111,7 @@ const formInput = WrappedComponent => {
           ref={element => {
             this.inputRef = element && element.querySelector("input");
           }}
+          onMouseDown={this.preventBlur}
```
alcferreira: Is this covered by storyshots? If not, we must add an unit test here...

packages/hoc-input/src/formInput.js
```diff
@@ -140,6 +142,12 @@ const formInput = WrappedComponent => {
       );
     }
 
+    componentDidMount() {
+      if (this.props.focused) {
+        this.setInputFocus();
```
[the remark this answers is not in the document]
alcferreira: IDK if the bug is on this part of the code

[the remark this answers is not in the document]
fgameiro: What's the expected behavior?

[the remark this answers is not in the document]
alcferreira: https://cravefood.github.io/farmblocks/?selectedKind=Search%20Field&selectedStory=update%20items&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

packages/search-field/src/SearchField.js
fgameiro: This boolean `autoFocus` is covered in the `with select item` in Storybook first as `undefined`, and if we clear the field, `autoFocus` returns true, then false. Is this enough?

Issue #438, opened 2018-07-11
Add Cancel/Save buttons to Protected input


alcferreira:
Reference https://projects.invisionapp.com/d/main#/console/13207780/308188844/preview

![image](https://user-images.githubusercontent.com/17936244/42594591-074f6fcc-8526-11e8-8dd9-b7773073a326.png)

fczuardi:
this will make the reuse of the protected text field by the search field harder...

fgameiro:
Small bug when trying to save something after saving an empty string:
![protected-input](https://user-images.githubusercontent.com/17071199/42634163-de9ac412-85b8-11e8-9249-24be579ea653.gif)

Steps:
1. edit the default value
2. save an empty string
3. edit
4. type something
5. cancel
6. edit
7. save without typing anything else (if you type, it will save)
8. it won't save

PR #439, opened 2018-07-11
feat(Protected input): add cancel and save buttons

alcferreira:
affects: @crave/farmblocks-input-text

ISSUES CLOSED: #438

## Closes Issues

* Closes #xxx

### Reviewer 1

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### Reviewer 2

* [x] Functions properly
* [ ] Has tests
* [ ] Has docs

### DIFF
--- a/packages/input-text/src/protectedValue.js
+++ b/packages/input-text/src/protectedValue.js
@@ -130,7 +130,7 @@ export default WrappedComponent => {
             {...wrappedComponentProps}
             protected={covered}
             focused={this.state.isEditing}
-            value={this.state.isEditing ? "" : this.state.value}
+            value={isEditing ? this.state.editedValue : this.state.value}
             onKeyDown={covered ? this.onKeyDown : onKeyDown}
             onChange={event => {
               this.setState({ editedValue: event.target.value });
@@ -141,7 +141,7 @@ export default WrappedComponent => {
               onBlur && onBlur(event);
             }}
           />
-          {covered && !this.state.isEditing && this._renderCover()}
+          {covered && !isEditing && this._renderCover()}
           {covered &&
             isEditing && (
               <div>
--- a/packages/input-text/src/protectedValue.js
+++ b/packages/input-text/src/protectedValue.js
@@ -50,31 +50,33 @@ export const protectedValueProps = {
 // until the edit button is clicked
 export default WrappedComponent => {
   return class ProtectedInput extends React.Component {
-    constructor(props) {
-      super(props);
-      this.state = {
-        isEditing: false,
-        value: props.value
-      };
-    }
+    state = {
+      isEditing: false,
+      value: this.props.value,
+      editedValue: ""
+    };
 
     onUncover = () => {
       this.setState({ isEditing: true });
     };
 
     onCancel = () => {
-      this.setState({ isEditing: false });
+      this.setState({ isEditing: false, editedValue: "" });
       return this.props.onCancel && this.props.onCancel();
     };
 
     onReplace = value => {
-      this.setState({ value, isEditing: false });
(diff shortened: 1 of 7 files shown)


packages/input-text/src/protectedValue.js
```diff
@@ -124,19 +126,37 @@ export default WrappedComponent => {
             focused={this.state.isEditing}
             value={this.state.isEditing ? "" : this.state.value}
             onKeyDown={covered ? this.onKeyDown : onKeyDown}
+            onChange={event =>
+              this.setState({ editedValue: event.target.value })
+            }
+            onBlur={() => {
+              this.state.isEditing && this.onCancel();
+            }}
```
fczuardi: this will override any onChange and onBlur properties passed to the input.

alcferreira: Sure thing... It broke some tests... Fixing

packages/input-text/src/protectedValue.js
```diff
@@ -97,19 +114,58 @@ export default WrappedComponent => {
...
+          {covered &&
+            isEditing && (
+              <div>
+                <Button
+                  id="cancel-button"
+                  onClick={this.onCancel}
+                  size={buttonSizes.MEDIUM}
+                  className="margin-button"
+                >
+                  Cancel
```
vNakamura: The `Cancel` and `Save` button labels should be customizable.  It's fine for now, given the urgency of this PR, but we'll definitely need this for localization.

Do you think it's quick enough to change now or should we create an issue to handle this later?

alcferreira: Good point, I'll do it now

fgameiro:
Small bug when trying to save something after saving an empty string:
protected-input

Steps:
1. edit the default value
2. save an empty string
3. edit
4. type something
5. cancel
6. edit
7. save without typing anything else (if you type, it will save)
8. it won't save

packages/input-text/src/protectedValue.js
```diff
@@ -23,43 +24,61 @@ const Cover = styled(styledInput)`
...
+
 // extra properties supported/extended by the protected field HOC
 export const protectedValueProps = {
   protected: PropTypes.bool,
   onReplace: PropTypes.func,
   onCancel: PropTypes.func,
-  onKeyDown: PropTypes.func
+  onKeyDown: PropTypes.func,
+  cancelButtonText: PropTypes.string,
+  saveButtonText: PropTypes.string
```
fczuardi: please document this new props on the README

packages/input-text/src/protectedValue.js
```diff
@@ -124,19 +126,37 @@ export default WrappedComponent => {
...
                 </Button>
-                <Button type={buttonTypes.SECONDARY} size={buttonSizes.MEDIUM}>
+                <Button
+                  id="save-button"
+                  onClick={() => {
+                    this.onReplace(this.state.editedValue);
+                  }}
+                  onMouseDown={e => {
+                    // We do this in order to avoid onBlur event on input
+                    e.preventDefault();
```
fczuardi: dont we want blur on save?

vNakamura: That's because it cancels the edit on blur. So clicking the save button will trigger the cancel(onBlur) before saving

alcferreira: we are preventing this blur https://github.com/CraveFood/farmblocks/pull/439/files#diff-4069488936b63491eb97f9c77d8a15baR140

fczuardi: ah, yes, ok

PR #441, opened 2018-07-16
Refactor search input

### LINKED ISSUE #433: Refactor text input on search component (to fix focus issues)
After #427 the text input have a mode that works like the input on search field, we should try to reuse / unify both inputs.

fczuardi:
## Closes Issues

* Closes #433 

### Reviewer 1

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### Reviewer 2

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs


packages/input-text/src/protectedValue.js
```diff
@@ -126,14 +130,18 @@ export default WrappedComponent => {
...
       return (
         <Container isEditing={isEditing}>
           <WrappedComponent
             {...wrappedComponentProps}
             protected={covered}
             focused={this.state.isEditing}
             value={isEditing ? this.state.editedValue : this.state.value}
-            onKeyDown={covered ? this.onKeyDown : onKeyDown}
+            onKeyDown={
+              covered && !disableManualReplace ? this.onKeyDown : onKeyDown
```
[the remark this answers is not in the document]
fczuardi: I can change the name, but that option also disable submitting with Enter and canceling with Escape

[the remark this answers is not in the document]
fczuardi: oh, yes, good catch

packages/search-field/src/SearchField.test.js
```diff
@@ -295,20 +292,42 @@ describe("SearchField", () => {
...
+    test("should reset highlight on Esc key", done => {
       event.key = "Escape";
       instance.onKeyDown(event);
       const stateChange = setStateSpy.mock.calls[0][0];
 
-      expect(stateChange).toHaveProperty("highlightedIndex", -1);
+      window.setTimeout(() => {
+        expect(stateChange).toHaveProperty("highlightedIndex", -1);
+        done();
+      }, 1000);
```
vNakamura: Is this setTimeout necessary?
`stateChange` is set before the timeout, so why wait 1 second  to check it's properties?

packages/search-field/src/SearchField.test.js
```diff
@@ -451,6 +461,12 @@ describe("SearchField", () => {
 
         expect(stateChange).toHaveProperty("highlightedIndex", -1);
       });
+
+      test("should reset inpuValue to previous selected item label", () => {
```
vNakamura: Typo: `inpuValue` -> `inputValue`

fczuardi:
requested changes addressed

packages/search-field/src/SearchField.js
```diff
@@ -38,17 +21,25 @@ class SearchField extends React.Component {
...
-    this.debouncedOnSearchChange(value);
-    if (!value) {
-      this.debouncedOnSearchChange.flush();
-      this.props.onChange();
+  setValueFromProps = () => {
+    const { value, items } = this.props;
+    const selectedItem = items && items.find(item => item.value === value);
+    if (selectedItem) {
+      this.setState({
+        inputValue: selectedItem.label,
```
vNakamura: `inputValue` should be cleared when `value` prop changes to `null` or `undefined`.

I'm not sure what should be the behavior when it receives a `value` which is not within the `items`.  But it should not keep the last label.
Maybe we could dispatch an error, so we can track it with Sentry? What you think?

fczuardi:
the label displacement is fixed:

search-css-fix-peek 2018-07-24 13-57

packages/search-field/src/SearchField.js
fczuardi: I am not sure either if we should throw, would an exception break the app?

fczuardi: I will just clear the input and the selection on both cases (update to no value, and update to non listed value)

PR #451, opened 2018-07-25
Select - add maxHeight prop

### LINKED ISSUE #450: Select - Add max height prop
This is needed to add scroll bars to select menu if we need to limit its height.

captura de tela de 2018-07-25 12-36-58

fgameiro:
## Closes Issues

* Closes #450 

### Reviewer 1

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### Reviewer 2

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### DIFF
--- a/packages/dropdown/src/styledComponents/DropdownMenuWrapper.js
+++ b/packages/dropdown/src/styledComponents/DropdownMenuWrapper.js
@@ -7,6 +7,9 @@ const DropdownMenuWrapper = styled.div`
   min-width: 100%;
   width: ${props => props.width};
   whitespace: no-wrap;
+  max-height: ${props => props.maxHeight};
+  overflow-y: auto;
+  scroll-behavior: smooth;
 
   border: solid 1px ${colorConstants.GREY_16};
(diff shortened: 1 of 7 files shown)


packages/dropdown/src/styledComponents/DropdownMenuWrapper.js
```diff
@@ -7,6 +7,9 @@ const DropdownMenuWrapper = styled.div`
   min-width: 100%;
   width: ${props => props.width};
   whitespace: no-wrap;
+  max-height: ${props => props.maxHeight}px;
```
alcferreira: receive `maxHeight` as string. We can set a percentage here as well.

vNakamura: Just to point out, a percentage won't work well here. The percentage would be relative to the components wrapper. Here's an example with `maxHeight='50%'`:
image

We need to use a viewport unit. `maxHeight='50vh':
image

PR #453, opened 2018-07-27
Fix select handling of value equal to zero

### LINKED ISSUE #452: Select - Incorrect handling when value is equal to zero
When we have to handle an item with `value = 0`, our currently select input doesn't get its label or get it wrong.

## Current behavior:
select-zero-error

## Expected behavior:
select-zero-fix

fgameiro:
## Closes Issues

* Closes #452 

### Reviewer 1

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### Reviewer 2

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs


packages/input-select/src/Select.js
```diff
@@ -62,14 +62,18 @@ class Select extends React.Component {
...
-    return item && item.label;
+    if (item) {
+      return item.label;
+    }
   };
 
   componentWillReceiveProps = newProps => {
     if (
-      (newProps.value &&
+      ((newProps.value || newProps.value === 0) &&
```
alcferreira: Add a unit test to cover this change

packages/input-select/src/Select.story.js
```diff
@@ -219,4 +219,56 @@ storiesOf("Select Input", module)
         maxHeight="150px"
       />
     ))
+  )
+  .add(
+    "Receiving new value",
```
alcferreira: For testing purposes, it'd be better to write unit tests...

packages/input-select/src/Select.js
fgameiro: done!

PR #466, opened 2018-08-01
Refactor selection bar

### LINKED ISSUE #462: Add Dropdown to Table's Selection Bar
Our current Table Selection Bar can have Primary and Secondary actions.
But we have a case now which requires 3 options:
image

- [ ] Needs #463

vNakamura:
## Closes Issues

* Closes #462

### Reviewer 1

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### Reviewer 2

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs


packages/table/src/SelectionBar.js
```diff
@@ -32,28 +22,9 @@ const SelectionBar = props => {
...
-              type={buttonTypes.PRIMARY}
-              onClick={buttonClick(selectedRows)(primaryAction)}
-            >
-              {props.primaryButtonLabel}
-            </Button>
-          </div>
-        )}
-      </div>
+      {children && (
+        <div className="actions">{children && children(selectedRows)}</div>
```
alcferreira: Double checking `children`... Is it really necessary?

packages/table/src/SelectionBar.js
```diff
@@ -1,25 +1,15 @@
...
-  const {
-    selectedRows,
-    clearSelection,
-    title,
-    primaryAction,
-    secondaryAction
-  } = props;
-  const hasActions = !!secondaryAction || !!primaryAction;
-  const hasSecondaryButton = !!secondaryAction;
+  const { selectedRows, clearSelection, height, title, children } = props;
```
fczuardi: probably ok, but do we really want to reuse the children prop for this? why use a property name that is normally a set of nodes to be a render function? will this break React.Children utilities (not that we use them, just curiosity)?

packages/table/src/Table.story.js
```diff
@@ -700,37 +700,71 @@ storiesOf("Table/SelectionBar", module)
...
+    "One action button",
     withInfo()(() => (
       <SelectionBar
         selectedRows={[{ name: "foo" }]}
         clearSelection={action("clear selection called")}
-        primaryAction={action("primary button clicked")}
-        primaryButtonLabel="Primary Action"
-      />
+      >
+        {selectedRows => (
```
fczuardi: this looks odd to me, is this a common pattern used by other libs? why not a property instead of a child?

alcferreira: We use the same concept on `Popover` component. The only diff is that there we pass a property called `content`

alcferreira: Hmmm... Yep, as we have access to the selectedRows outside the `SelectionBar` I agree with you... It makes sense to render children normally

packages/table/src/SelectionBar.js
fczuardi: Google searching this pattern of function as child, there are results for and against it, but the main argument of [this one here, against it](https://americanexpress.io/faccs-are-an-antipattern/) is that by using this pattern we lose the ability of having meaningful names, and that we are re-purposing a familiar prop for the sake of clever hackism. I tend to agree with that conclusions.

packages/table/src/Table.story.js
alcferreira: @vNakamura just tested with regular children and it works fine... Let's change that?!

alcferreira:
There's a left over on story `With selectable rows and a selection header`:
```
          <SelectionBar
            selectedRows={selectedRows}
            clearSelection={clearFunction}
            title={count =>
              count === 1 ? "1 fruit selected" : `${count} Fruits selected`
            }
            primaryButtonLabel="Connect"
            primaryAction={action("primary button clicked")}
            secondaryButtonLabel="View Profile"
            secondaryAction={action("secondary button clicked")}
          />
```

packages/table/src/Table.story.js
```diff
@@ -187,7 +187,24 @@ storiesOf("Table/Basic", module)
             primaryAction={action("primary button clicked")}
             secondaryButtonLabel="View Profile"
             secondaryAction={action("secondary button clicked")}
-          />
+          >
```
alcferreira: remove old properties above:
```
primaryButtonLabel
primaryAction
secondaryButtonLabel
secondaryAction
```

packages/table/src/Table.story.js
```diff
@@ -187,7 +187,24 @@ storiesOf("Table/Basic", module)
...
+            <div
+              style={{ display: "grid", gridGap: 8, gridAutoFlow: "column" }}
+            >
+              <Button
+                type={buttonTypes.SECONDARY}
+                onClick={() => action("secondary button clicked")()}
```
alcferreira: Currently our story logs the `selectedRows` as well... Maybe we should keep logging it .

No big deal, change if you want to, but this could be written like this: `onClick={action("secondary button clicked")}`

vNakamura: I don't get it.
The `children` is `PropTypes.node`. We can't have `selectedRows` anymore.

packages/table/src/Table.story.js
vNakamura: Oops!

packages/table/src/Table.story.js
alcferreira: We do have access to `selectedRows` (:

```
               <Button
                  type={buttonTypes.SECONDARY}
                  onClick={() =>
                    action("secondary button clicked")(selectedRows)
                  }
                >
```

image

vNakamura: Oh, ok! We're talking about the story for the `Table` with `SelectionHeader`.
I tought it was the story for `SelectionHeader` alone.

Fixed!

packages/table/src/styledComponents/SelectionBar.js
```diff
@@ -1,13 +1,11 @@
...
 import { colors } from "@crave/farmblocks-theme";
 
-const height = "64px";
-
 const SelectionBar = styled.div`
   display: flex;
   justify-content: right;
   align-items: center;
-  height: ${height};
+  height: ${props => props.height};
```
fczuardi: please update the readme to include this property

PR #470, opened 2018-08-06
Fix Search Field inputValue when items change after mount

### LINKED ISSUE #469: Search Field - inputValue is not updated when items change after mount
When, after mount, `items` change, and `value` is the same, we don't update the state `inputValue`, therefore we may show incorrect value on Search Field.

The following sequence illustrates this problem happening on webapp:
inputvalue

fgameiro:
## Closes Issues

* Closes #469 
### Reviewer 1

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### Reviewer 2

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### DIFF
--- a/packages/search-field/src/SearchField.js
+++ b/packages/search-field/src/SearchField.js
@@ -65,6 +65,7 @@ class SearchField extends React.Component {
 
     if (!isEqual(prevProps.items, this.props.items)) {
       this.setState({ items: this.props.items });
+      this.setValueFromProps();
     }
 
     if (prevProps.value !== this.props.value) {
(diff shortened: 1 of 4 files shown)


packages/search-field/src/SearchField.js
```diff
@@ -65,6 +65,7 @@ class SearchField extends React.Component {
 
     if (!isEqual(prevProps.items, this.props.items)) {
       this.setState({ items: this.props.items });
+      this.setValueFromProps();
```
alcferreira: I guess we should have an unit test for this because snapshot by itself won't cover this case.

fgameiro: Added!

PR #481, opened 2018-08-10
Remove default tooltip from disabled components

### LINKED ISSUE #478: Remove disabled tooltip
As the disabled tooltip is not being used anywhere, except by the amount-selectors, we should remove it from the following components:
- [x] input-checkbox
- [x] input-radio
- [x] input-select
- [x] input-text
- [x] link

On button component, pass `noTooltip = true` as default. Tooltip should be visible only if there's a tooltipText passed to the button component.

All changes are breaking changes.

fgameiro:
## Closes Issues

* Closes #478 

### Reviewer 1

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### Reviewer 2

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs


packages/input-checkbox/src/Checkbox.js
```diff
@@ -18,20 +13,8 @@ const createCheckbox = ({ isSwitch }) =>
     };
 
     render() {
-      const {
-        type,
```
alcferreira: `type` should still be removed from `inputProps`. Otherwise the `type` could be wrongly passed to `input`

packages/input-checkbox/src/Checkbox.js
```diff
@@ -40,32 +23,25 @@ const createCheckbox = ({ isSwitch }) =>
...
-            <div className="visibleCheckbox">
-              <div className={isSwitch ? "toggle" : "checkmark wg-check"} />
-            </div>
-            {label && (
-              <Text title={!isSwitch} type={fontColor} size={fontSizes.MEDIUM}>
-                {label}
-              </Text>
-            )}
-          </TooltipTarget>
+          <div className="visibleCheckbox">
```
alcferreira: Is this expected?

packages/input-radio/src/Radio.js
```diff
@@ -53,25 +44,24 @@ class Radio extends React.Component {
...
 
-    const fontColor =
-      (inputProps.disabled && fontTypes.SUBTLE) || fontTypes.NORMAL;
+    const fontColor = (disabled && fontTypes.SUBTLE) || fontTypes.NORMAL;
 
     return (
       <Label {...labelProps}>
-        <TooltipTarget {...tooltipProps} className="tooltipTarget">
-          <input className="hiddenInput" type="radio" {...inputProps} />
+        <div className="tooltipTarget">
```
alcferreira: Change `tooltipTarget` class name to another one that reflects its style.

packages/input-radio/src/Radio.js
```diff
@@ -53,25 +44,24 @@ class Radio extends React.Component {
...
+          <input
+            disabled
```
alcferreira: Remove this

packages/input-checkbox/src/Checkbox.js
```diff
@@ -18,20 +13,8 @@ const createCheckbox = ({ isSwitch }) =>
...
-        onChange,
-        displayBlock,
-        onMouseLeave,
-        onMouseOver,
-        tooltipText,
-        tooltipAlign,
-        noTooltip,
-        ...inputProps
-      } = this.props;
+      const { label, checked, onMouseUp, onChange, ...inputProps } = this.props;
```
fczuardi: `displayBlock` is not related with tooltip, I think we still need it

packages/input-select/src/Select.js
```diff
@@ -19,20 +16,16 @@ import Item from "./components/Item";
...
-)("input");
+const EnhancedInput = compose(withMessages, formInput, withImage)("input");
 EnhancedInput.displayName = "EnhancedInput";
 
 class Select extends React.Component {
   render() {
+    const { width, zIndex, disabled, items } = this.props;
     return (
-      <DropdownWrapper width={this.props.width} zIndex={this.props.zIndex}>
+      <DropdownWrapper width={width} zIndex={zIndex} disabled={disabled}>
```
alcferreira: Remove `disabled`. This property actually has no effect on `DropdownWrapper`.

packages/input-select/src/Select.js
```diff
@@ -190,9 +182,9 @@ class Select extends React.Component {
     disableSearch: PropTypes.bool,
     zIndex: PropTypes.number,
     maxHeight: PropTypes.string,
+    disabled: PropTypes.bool,
```
alcferreira: Remove it... This property is documented on `formInputProps`

packages/input-text/src/TextInput.js
```diff
@@ -32,27 +24,26 @@ const TextInput = props => {
...
   ...formInputProps,
   ...protectedValueProps,
   ...withMessagesProps,
 
   type: PropTypes.string,
   placeholder: PropTypes.string,
   required: PropTypes.bool,
-  mask: PropTypes.string
+  mask: PropTypes.string,
+  disabled: PropTypes.bool
```
alcferreira: Remove it... This property is documented on `formInputProps`

packages/button/src/Button.js
```diff
@@ -49,7 +56,7 @@ const Button = props => {
 Button.defaultProps = {
   size: buttonSizes.SMALL,
   type: buttonTypes.NEUTRAL,
-  tooltipText: "This action is disabled"
+  noTooltip: true
```
alcferreira: Unused property.

packages/input-checkbox/src/Checkbox.js
fgameiro: It wasn't intentional, but this is a effect of removing the `tooltipTarget` div that became useless

packages/input-checkbox/src/Checkbox.js
fczuardi: nevermind, lets remove displayBlock

packages/input-checkbox/src/Checkbox.js
alcferreira: `const { label, checked, onMouseUp, onChange, ...inputProps } = this.props;`

`<Checkbox type='number' />` will pass `type` to `<input type="checkbox" {...inputProps} className="hiddenCheckbox" />`

packages/input-checkbox/src/Checkbox.js
alcferreira: All the other components remained the same... weird

packages/input-checkbox/src/Checkbox.js
fczuardi: everything that is not declared before the spread is part of `inputPropos`

packages/input-text/src/TextInput.story.js
```diff
@@ -107,17 +107,6 @@ storiesOf("Text Input", module)
...
-    "Disabled with custom Tooltip text",
-    withInfo()(() => (
-      <TextInput
-        tooltipText="a custom tooltip"
-        disabled
-        label="Field Label"
-        placeholder="input value here"
-      />
-    ))
-  )
```
fczuardi: there is still a `tooltipText` 7 lines bellow that can be removed :point_down:

packages/input-checkbox/src/Checkbox.js
fczuardi: I am in favor of removing that type however, because it doesnt make much sense to pass a type attribute to a Checkbox component, we don't have instances of such combo neither in webapp nor farmblocks, and may not happen (specially considering this prop is not documented). It is different from the case of the input text component, that someone could want to forward the type="email" type="phone" or something like that to the underlining html input, in this case the underlining html input will be always type="checkbox"

alcferreira: Just to document what we talked about.
Let's just change the order of the properties from 
`<input type="checkbox" {...inputProps}  className="hiddenCheckbox" />` 
to 
`<input {...inputProps} type="checkbox" className="hiddenCheckbox" />`

packages/input-checkbox/src/Checkbox.js
alcferreira: Version 2.0.0:
image

After changes:
image

PR #494, opened 2018-08-15
(Select) Highlight input text on menu open

### LINKED ISSUE #491: [Select] Highlight input text on menu open
When we have an item already selected and we open the menu, the input for search gets focus but the cursor is at the end of the text.

It would be better if the text was highlighted/selected so the user could start typing a new option without having to delete the current text

fgameiro:
## Closes Issues

* Closes #491 

### Reviewer 1

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### Reviewer 2

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### DIFF
--- a/packages/input-select/src/Select.js
+++ b/packages/input-select/src/Select.js
@@ -35,12 +35,20 @@ class Select extends React.Component {
           renderMenu={this._renderMenu}
           renderItem={this._renderItem}
           autoHighlight={false}
-          onMenuVisibilityChange={isMenuOpen => this.setState({ isMenuOpen })}
+          onMenuVisibilityChange={this.onMenuVisibilityChange}
           wrapperStyle={{}}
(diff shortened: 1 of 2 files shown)


packages/input-select/src/Select.js
```diff
@@ -35,13 +35,21 @@ class Select extends React.Component {
...
           wrapperStyle={{}}
+          ref={ref => (this.input = ref)}
         />
       </DropdownWrapper>
     );
   }
 
+  onMenuVisibilityChange = isMenuOpen => {
+    if (!this.props.disableSearch && isMenuOpen) {
+      this.input.select();
```
vNakamura: Add `&& this.input` to the if condition. Just to prevent an error if the ref is not set yet.

fgameiro: done!

PR #498, opened 2018-08-16
[Table] Fix alignments in Table row groups

### LINKED ISSUE #495: Fix alignments in Table row groups
- __Blue thing `⌞` on the left__: Align vertically at center when row is taller than normal
- __Expand button__: Align right or prevent it's column from expanding when table has free space

image

fgameiro:
## Closes Issues

* Closes #495 

### Reviewer 1

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### Reviewer 2

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### DIFF
--- a/packages/table/src/Table.js
+++ b/packages/table/src/Table.js
@@ -185,7 +185,7 @@ class Table extends React.Component {
             })
         )}
         {collapsed && (
-          <BodyCell className="cell" {...rowProps}>
+          <BodyCell className="cell" align="right" {...rowProps}>
             {group && this._renderExpandToggle(rowKey)}
           </BodyCell>
         )}
--- a/packages/table/src/Table.story.js
+++ b/packages/table/src/Table.story.js
@@ -587,6 +587,17 @@ storiesOf("Table/Row Groups", module)
     })
   )
   .add(
+    "Expandable Groups - Wide table",
+    withInfo()(() => {
+      return (
+        <Table width="800px" data={orders} collapsed rowGroupKey="suborders">
+          <Column title="Name" text={row => row.name} />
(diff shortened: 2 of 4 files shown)


packages/table/src/Table.js
```diff
@@ -219,13 +220,15 @@ class Table extends React.Component {
...
       : "wg-small-arrow-bottom";
     return (
-      <Button
-        icon={icon}
-        onClick={e => {
-          this.expandToggleClicked(rowKey);
-          e.stopPropagation();
-        }}
-      />
+      <ButtonContainer>
```
alcferreira: `ButtonContainer` is a `div` with flex and justify-content only.
I think it'd be better to set these styles on `BodyCell` component.

This way we would have this:

```
<div className="class-name-x">
  <Button
    icon={icon}
    onClick={e => {
      this.expandToggleClicked(rowKey);
      e.stopPropagation();
    }}
  />
</div>
```

And `BodyCell` would be like this:
```
export const BodyCell = styled.td`
  text-align: ${textAlign};
  background-color: ${cellBg};

  .class-name-x {
    display: flex;
    justify-content: flex-end;
  }
`;
```

alcferreira:
Please set a width to some story so we can easily see that the expand button is right-aligned.
I had to do it manually to check the changes...

packages/table/src/Table.js
vNakamura: From this comment I noticed that we have this `text-align` rule already, which can be set by the `align` prop. So you can simple use this at line 188: 
```
<BodyCell className="cell" align="right" {...rowProps}>
```

PR #508, opened 2018-08-23
Interactive Map Balloon

### LINKED ISSUE #506: Interactive Map Balloon
https://projects.invisionapp.com/d/main#/console/15082834/313631252/preview

image

- [x]  ~~Add `interactive` prop (bool)~~
- [x]  Add prop `onPinClick`, so we can open the ballon from the app. Maybe add an `autoExpand` option, so it will open automatically, but I don’t think it’s necessary.
- [x]  Add prop  `onBalloonClick`
- [x] Receive a reference prop that will be passed to the `onPinClick` and `onBalloonClick` to help identify which pin/balloon was clicked
- [x]  Add `customPinColor` and `customPinHighlightColor` props
- [x] Disable text selection on touch move [CSS user-select](https://developer.mozilla.org/en-US/docs/Web/CSS/user-select)
- [x] Visual changes:
  - [x] Circle around pin (only for interactive)
  - [x] Icon `>` indication for interactive balloons

vNakamura:
affects: @crave/farmblocks-map-balloon

## Closes Issues

* Closes #506

### Reviewer 1

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### Reviewer 2

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### DIFF
--- a/packages/map-balloon/README.md
+++ b/packages/map-balloon/README.md
@@ -78,10 +78,14 @@ farm symbol, and that the class name to include that icon is `.wg-purveyor`.
 * `balloonSize`: number (default: 260). Balloon size
 * `borderRadius`: string (default: 8px). Defines the balloon border radius
 * `pinColor`: string (default: #f1c618). Color to be set to the pin icon
+* `pinHighlightColor`: string (defaults to the `pinColor` value). Color to be set to the highlighted pin
 * `pinSize`: number (default: 40). Size to be set to the pin icon
 * `opacity`: number (default: 1)
 * `captionSize`: number (default: 18), Size of the caption text.
 * `imageTextSize`: number (default: 28), Size of the text overlay on images.
+* `onPinClick`: function. Called when the pin is clicked. If set, changes the pin visual into a round button
+* `onBalloonClick`: function. Called when the balloon is clicked. If set, adds an indicator to the balloon text.
+* `reference`: any. A value that will be passed as first argument for `onPinClick` and `onBalloonClick` calls. This can be handy to identify which pin/balloon was clicked.
 
 ## License
 
--- a/packages/map-balloon/README.md
+++ b/packages/map-balloon/README.md
@@ -85,7 +85,7 @@ farm symbol, and that the class name to include that icon is `.wg-purveyor`.
 * `imageTextSize`: number (default: 28), Size of the text overlay on images.
 * `onPinClick`: function. Called when the pin is clicked. If set, changes the pin visual into a round button
 * `onBalloonClick`: function. Called when the balloon is clicked. If set, adds an indicator to the balloon text.
-* `reference`: any. A value that will be passed as first argument for `onPinClick` and `onBalloonClick` calls. This can be handy to identify which pin/balloon was clicked.
+* `value`: any. A value that will be passed as first argument for `onPinClick` and `onBalloonClick` calls. This can be handy to identify which pin/balloon was clicked.
 
 ## License
 
--- a/packages/map-balloon/src/MapBalloon.js
+++ b/packages/map-balloon/src/MapBalloon.js
@@ -4,6 +4,7 @@ import values from "object.values";
 import Text, { fontTypes } from "@crave/farmblocks-text";
 import { colors, fontSizes } from "@crave/farmblocks-theme";
 import { Flex, Box } from "grid-styled";
+import { CSSTransition } from "react-transition-group";
 
 import alignments from "./constants/alignments";
 import Wrapper from "./styledComponents/Wrapper";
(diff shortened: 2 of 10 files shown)


packages/map-balloon/src/MapBalloon.js
```diff
@@ -22,44 +24,85 @@ const MapBalloon = ({
...
+        <div>
+          <Pin
+            className={`${pinIconClass} ${open ? "open" : ""}`}
+            pinColor={pinColor}
+            pinHighlightColor={pinHighlightColor}
+            pinSize={pinSize}
+            interactive={interactivePin}
+            animated={animated}
+            onClick={
+              interactivePin ? event => onPinClick(reference, event) : null
```
alcferreira: Any special reason to pass `null` to onClick? Just curious (:

This could be written like this as well `onClick={event => onPinClick && onPinClick(reference, event}`

vNakamura: No particular reason. I first tried `onClick={interactivePin && (event => onPinClick(reference, event))}` but this resulted in some proptype warnings when `interactivePin` was `false`. So I just changed to a ternary with `null`.

Your suggestion works with no warnings. My only concern is that we're creating a new function at every render of every balloon, even it doesn't have an `onPinClick` set.

What about `onClick={onPinClick && (event => onPinClick(reference, event))}`?

packages/map-balloon/src/MapBalloon.js
```diff
@@ -22,44 +24,85 @@ const MapBalloon = ({
...
   balloonSize,
   captionSize,
-  imageTextSize
+  imageTextSize,
+  reference,
+  onPinClick,
+  onBalloonClick
 }) => {
+  const interactivePin = !!onPinClick;
+  const interactiveBalloon = !!onBalloonClick;
```
fczuardi: :+1:

packages/map-balloon/README.md
```diff
@@ -78,10 +78,14 @@ farm symbol, and that the class name to include that icon is `.wg-purveyor`.
...
 * `borderRadius`: string (default: 8px). Defines the balloon border radius
 * `pinColor`: string (default: #f1c618). Color to be set to the pin icon
+* `pinHighlightColor`: string (defaults to the `pinColor` value). Color to be set to the highlighted pin
 * `pinSize`: number (default: 40). Size to be set to the pin icon
 * `opacity`: number (default: 1)
 * `captionSize`: number (default: 18), Size of the caption text.
 * `imageTextSize`: number (default: 28), Size of the text overlay on images.
+* `onPinClick`: function. Called when the pin is clicked. If set, changes the pin visual into a round button
+* `onBalloonClick`: function. Called when the balloon is clicked. If set, adds an indicator to the balloon text.
+* `reference`: any. A value that will be passed as first argument for `onPinClick` and `onBalloonClick` calls. This can be handy to identify which pin/balloon was clicked.
```
fczuardi: `any` looks very flexible, in practice, what will be normally used here? a string? a dom node? a react node? Are we using this property for more than a type or we are always using with the same type?

packages/map-balloon/src/MapBalloon.js
```diff
@@ -3,6 +3,8 @@ import PropTypes from "prop-types";
 import values from "object.values";
 import Text, { fontTypes } from "@crave/farmblocks-text";
 import { colors, fontSizes } from "@crave/farmblocks-theme";
+import { Flex, Box } from "grid-styled";
```
fczuardi: What advantages bringing in a grid layout framework will give us? Craity / readabality? Compatibility? Future-proofness? It looks like the demands for a layout system inside a balloon currently is not much today, is the trade-of in styling declaratively with components vs using plain css good for us? Is the time for this right? I dont have the answer, just curious...:)

packages/map-balloon/src/MapBalloon.story.js
```diff
@@ -170,4 +188,76 @@ storiesOf("Map Balloon", module)
         />
       </DemoGrid>
     ))
+  )
+  .add(
+    "Interactive",
+    withInfo()(() => (
+      <DemoGrid>
+        <MapBalloon
+          interactive
```
fczuardi: this flag is no longer necessary, right?

packages/map-balloon/README.md
vNakamura: `reference` is never used by this component. It only stores it an send it back at `onPinClick` and `onBalloonClick`. So the type is irrelevant for this component.

At the webmap we use it to pass the producer id (number), but I don't see why we should restrict this proptype.

fczuardi: Hmm, webmap could bind this producer id onto the callback and have it executed without arguments then...

Or we could rename `reference` to `value`, like how radio buttons do. (yeah, maybe what bothers me is the prop name)

Or we could pass the producer id as a `data-something` attribute and have the click handlers forward the click event.

But leaving as it is now is not a problem.

packages/map-balloon/src/MapBalloon.js
vNakamura: Yeah, it's not really necessary.

vNakamura:
We need to fix lerna before merging this

PR #574, opened 2018-09-27
[formInput] added Autofocus property 

### LINKED ISSUE #573: [formInput] prevent auto focus styles
We have a situation where we need to keep the focused style on an input even though it has lost focus.

alcferreira:
## Closes Issues

* Closes #573

### Reviewer 1

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### Reviewer 2

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs


packages/hoc-input/src/formInput.js
```diff
@@ -35,7 +35,8 @@ export const formInputProps = {
   moreInfoContent: PropTypes.node,
   moreInfoAlign: PropTypes.oneOf(values(alignments)),
   prefix: PropTypes.string,
-  suffix: PropTypes.string
+  suffix: PropTypes.string,
+  autoFocus: PropTypes.bool
```
vNakamura: The name of the prop is dubious. It sounds like it would focus the input on mount.

How about `autoFocusHighlight` or `autoFocusStyled`? I'm not 100% about these, but I couldn't think of anything better.

alcferreira: Maybe `autoControlFocusedStyle`?

vNakamura: OK, sound good!

PR #591, opened 2018-10-16
[WIP] Add custom properties to tag component

### LINKED ISSUE #586: [Tag] Adapt to use as delivery badge on webapp
- [ ] Create a `small` type with:
   - `height`: `16px`
   - `font-size`: `8px`
- [ ] Add a property `icon`
- [ ] Add custom background color
- [ ] Add custom foreground color

fgameiro:
## Closes Issues

* Closes #586 

### Reviewer 1

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### Reviewer 2

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs


packages/tag/src/components/Tag.js
```diff
@@ -1,33 +1,56 @@
...
   return (
-    <StyledTag {...props}>
-      {text}
-      {onRemove && (
-        <div className="icon" onClick={() => onRemove(value)}>
-          <i className="wg-close" />
-        </div>
-      )}
-    </StyledTag>
+    <ThemeProvider theme={style}>
```
alcferreira: We should wrap a `Tag` component with `ThemeProvider` outside the `Tag` component.
Check the docs: https://www.styled-components.com/docs/advanced#theming

We will style normally the `Tag` component. The difference is that we get the `type` from the `theme` provided. That is if no theme was provided, the component will render with the default style. If a theme is provided, we use the properties that come from the custom theme.

fgameiro:
This PR was originally intended to add props to allow more styling options, but after some conversations on Slack we decided handle only the icon addition. 
Because of this I am changing the current title, formerly `Add custom properties to tag component`.

packages/tag/src/components/styledComponent/StyledTag.js
```diff
@@ -9,18 +9,25 @@ const typeStyles = {
   [tagTypes.SECONDARY]: css`
     background-color: ${colorConstants.INDIGO_MILK_CAP};
     color: white;
+    .icon {
+      color: white;
```
alcferreira: We don't need to set `color` again if the color is the same used on the parent node... please remove it here and on line 29

PR #600, opened 2018-10-18
[Image] add css property

### LINKED ISSUE #598: [Image] add opacity property
We need a way to set `opacity` on Image component

alcferreira:
## Closes Issues

* Closes #598

### Reviewer 1

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### Reviewer 2

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### DIFF
--- a/packages/image/src/Image.js
+++ b/packages/image/src/Image.js
@@ -20,7 +20,8 @@ Image.propTypes = {
   }),
   width: PropTypes.string,
   height: PropTypes.string,
-  fit: PropTypes.string
+  fit: PropTypes.string,
+  css: PropTypes.object
 };
 
 Image.defaultProps = {
--- a/packages/image/src/Image.js
+++ b/packages/image/src/Image.js
@@ -21,7 +21,11 @@ Image.propTypes = {
   width: PropTypes.string,
   height: PropTypes.string,
   fit: PropTypes.string,
-  css: PropTypes.object
+  css: PropTypes.oneOfType([
+    PropTypes.object,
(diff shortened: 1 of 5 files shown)


vNakamura:
Isn't this doing the same thing `style` already do?

alcferreira:
Yep... I just want to avoid using the `style` tag

vNakamura:
> Yep... I just want to avoid using the `style` tag

I don't get it. Any problem with `style` that I'm unaware of?

alcferreira:
> > Yep... I just want to avoid using the `style` tag
> 
> I don't get it. Any problem with `style` that I'm unaware of?

Not necessary a problem, but it's a good practice to avoid using style
https://reactjs.org/docs/dom-elements.html#style

> Are inline styles bad?
CSS classes are generally better for performance than inline styles.

packages/image/src/Image.js
```diff
@@ -20,7 +20,8 @@ Image.propTypes = {
   }),
   width: PropTypes.string,
   height: PropTypes.string,
-  fit: PropTypes.string
+  fit: PropTypes.string,
+  css: PropTypes.object
```
vNakamura: `styled-components` also works with `string` or with the ``css` `;`` helper (which returns an `array`).
Let's accept those as well.

```suggestion
  css: PropTypes.oneOfType([
    PropTypes.object,
    PropTypes.string,
    PropTypes.array
  ])
```

PR #619, opened 2018-10-29
Fix invisible tooltip

### LINKED ISSUE #618: Tooltip is over components even when it's invisible
The Tooltip component is mounted even when it's invisible, which is causing it to be over other components.

Example:
order_options
screen shot 2018-10-29 at 12 32 19

fgameiro:
## Closes Issues

* Closes #618 

### Reviewer 1

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### Reviewer 2

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs


fgameiro:
@alcferreira @vNakamura please take a look when you can :)

packages/hoc-withtooltip/src/withTooltip.js
```diff
@@ -4,6 +4,7 @@ import wrapDisplayName from "recompose/wrapDisplayName";
 import PropTypes from "prop-types";
 import styled from "styled-components";
 import values from "object.values";
+import ReactCSSTransitionGroup from "react-addons-css-transition-group";
```
vNakamura: This is an outdated version. Use this one: https://reactcommunity.org/react-transition-group/

Also, it should be in the package's dependencies list. Go to the package folder and use `yarn add react-transition-group` there.

alcferreira: We also should remove the animations set on tooltip component:
https://github.com/CraveFood/farmblocks/blob/dabb0163321f0cb7af087835f9d591cc2f2db9d9/packages/tooltip/src/styledComponents/Tooltip.js#L80

PR #707, opened 2019-02-22
Fix item id creation

### LINKED ISSUE #706: [Select] - Fix Items ID on if the item label has whitespace
## Expected behavior
The items IDs are based on item label, if the label string has whitespaces remove it to generate the ID. 

## Current behavior
If the item label string has whitespaces it is passed to generate the ID too.

MateusStanki:
## Closes Issues

* Closes #706

### DIFF
--- a/packages/input-select/src/Select.js
+++ b/packages/input-select/src/Select.js
@@ -148,7 +148,7 @@ class Select extends React.Component {
         ) : (
           <Item
             label={item.label}
-            id={this.props.id && `${this.props.id}-item-${item.label}`}
+            id={this.props.id && `${this.props.id}-item-${item.value}`}
             image={item.image}
             selected={selected}
           />


packages/input-select/src/Select.js
```diff
@@ -148,7 +148,10 @@ class Select extends React.Component {
         ) : (
           <Item
             label={item.label}
-            id={this.props.id && `${this.props.id}-item-${item.label}`}
+            id={
+              this.props.id &&
+              `${this.props.id}-item-${item.label.replace(/\s/g, "")}`
```
vNakamura: How about we use `item.value`? It's less probable of having spaces, so we could simply use it here  without the `replace`.

This component already has some performance issues. Once I used it for a country selector and it got janky with about 200 items. Using regex in every item render could make it even less performant.

I'm usually against premature optimization, but in this case, as it will be used by QA, if we decide to change it later, it will break their tests.

MateusStanki: Since I am documenting all created ids, I agree, it's better to be by value.

PR #712, opened 2019-02-27
Configure prettier linter and airbnb linter rules

### LINKED ISSUE #698: Fix inconsistency of line width among devs
Usually it happens on pre-commit. Prettier/ESLint will auto fix some files adding breaks in long lines or arrays. E.g. changing this:
```js
  "files": ["AUTHORS", "LICENSE", "README.md", "CHANGELOG.md", "lib"],
```
To this:
```js
  "files": [
    "AUTHORS",	
    "LICENSE",	
    "README.md",	
    "CHANGELOG.md",	
    "lib"	
  ],
```

Then later, running from another machine, it will change it back to a single line. This adds noise to the PRs.

alcferreira:
## Closes Issues

* Closes #698

### Reviewer 1

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### Reviewer 2

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs


packages/link/src/components/Link.story.js
```diff
@@ -2,11 +2,10 @@ import React from "react";
 import { storiesOf } from "@storybook/react";
 import { action } from "@storybook/addon-actions";
 
+import { fontSizes } from "@crave/farmblocks-theme";
```
fgameiro: We usually would group and move this kind of import to line 4. Any reason to place it here?

alcferreira: This line was auto fixed by eslint

packages/tag/src/components/Tag.js
```diff
@@ -16,7 +16,13 @@ const Tag = props => {
       )}
       {text || children}
       {onRemove && (
-        <div className="close-icon" onClick={() => onRemove(value)}>
+        <div
+          role="button"
+          tabIndex="0"
+          className="close-icon"
+          onClick={() => onRemove(value)}
+          onKeyDown={() => onRemove(value)}
```
fgameiro: It's kinda weird allow any key pressing to call `onRemove`, doesn't it? We could restrict some keys (`enter` and `space`) to call the action...

packages/table/src/constants.js
```diff
@@ -1,4 +1,6 @@
 export const rowHeights = {
   SMALL: "64px",
-  MEDIUM: "80px"
+  MEDIUM: "80px",
 };
+
+export default rowHeights;
```
vNakamura: If we're gonna export `rowHeights` as default, we should rename the file.

I think this file was supposed to have other types of constants. So, how about we export an object containing `rowHeights`? This way we won't have breaking changes if we need to add another group of constants in the future.
```suggestion
export default { rowHeights };
```

vNakamura:
`Dropdown/Component/WithRef` story is not logging the action anymore

vNakamura:
`Select Input` is rendering `false` instead of the placeholder:
image

master:
image

vNakamura:
Label component always looks as default

packages/tag/src/components/Tag.js
vNakamura: Yep, even `tab` is triggering `onRemove`. So a user can't navigate with keyboard through the tags without removing them.

packages/button/src/Button.js
```diff
@@ -10,70 +10,73 @@ import StyledButton from "./styledComponents/Button";
...
-  icon,
-  rightIcon,
-  text,
-  children,
-  noTooltip,
-  ...buttonProps
-}) => {
-  // @FIXME revisit font icon assets on farmblocks
-  const iconName = buttonProps.loading ? "wg-loading" : icon;
+// eslint-disable-next-line
```
vNakamura: Can we disable just the rule that's affecting us?
Also, let's add a comment on why it had to be disabled as it's not obvious from the code.

packages/input-select/src/Select.js
```diff
@@ -70,6 +70,7 @@ class Select extends React.Component {
     this.props.onChange(item.value);
   };
 
+  // eslint-disable-next-line
```
vNakamura: It's `consistent-return` that is throwing an error, right?
Let's disable just that.

packages/button/src/Button.js
alcferreira: Sure!

PR #733, opened 2019-03-15
Multi select

### LINKED ISSUE #355: Select - duplicate check mark appearing on itens with same label
When a item is selected and it happens to have the same label of other item, a check mark appears in both of them, even if they are different (i.e. have different ids, or values). 

captura de tela de 2018-05-14 21-46-17

### LINKED ISSUE #721: Create Multi select component
Specs: https://app.goabstract.com/projects/23793d20-e68f-11e8-816f-b177d180e919/branches/master/commits/fc29630f7f467aa0aae2035a2f50119faef49218/files/A1233ED6-A996-4D7E-A77E-11A55CED1624/layers/3D7225DB-4979-41AB-ACD8-676166F5E8B4?mode=build

image

vNakamura:
## Closes Issues

* Closes #721
* Closes #355

### Reviewer 1

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### Reviewer 2

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### DIFF
--- a/packages/input-select/src/Select.js
+++ b/packages/input-select/src/Select.js
@@ -11,17 +11,19 @@ import {
 import withMessages, {
   withMessagesProps,
 } from "@crave/farmblocks-hoc-validation-messages";
+import Tag from "@crave/farmblocks-tags";
 
 import withImage, { refName } from "./components/withImage";
 import Item from "./components/Item";
 import EmptyCard from "./components/EmptyCard";
 import DropdownWrapper from "./styledComponents/DropdownWrapper";
+import InputWithTags from "./components/InputWithTags";
 
 const EnhancedInput = compose(
   withMessages,
   formInput,
   withImage,
-)("input");
+)(InputWithTags);
 EnhancedInput.displayName = "EnhancedInput";
 
 class Select extends React.Component {
@@ -79,6 +81,8 @@ class Select extends React.Component {
     onChange(item.value);
   };
 
+  onRemoveTag = value => this.onSelect("", { value });
+
   // eslint-disable-next-line consistent-return
   getSelectedLabel = props => {
     const item =
@@ -119,7 +123,20 @@ class Select extends React.Component {
         innerRef={ref}
         refName={refName}
         image={image}
-      />
+      >
+        {multi &&
+          items.map(
+            item =>
+              this.state.selectedValue.includes(item.value) && (
+                <Tag
+                  key={item.value}
+                  value={item.value}
+                  text={item.label}
+                  onRemove={this.onRemoveTag}
+                />
+              ),
+          )}
+      </EnhancedInput>
     );
   };
 
--- a/packages/input-select/src/Select.js
+++ b/packages/input-select/src/Select.js
@@ -142,6 +142,7 @@ class Select extends React.Component {
             id={this.props.id && `${this.props.id}-item-${item.value}`}
             image={item.image}
             selected={selected}
+            checkbox={this.props.multi}
           />
         )}
       </DropdownItemWrapper>
@@ -209,6 +210,7 @@ class Select extends React.Component {
     disableSearch: PropTypes.bool,
     zIndex: PropTypes.number,
     maxHeight: PropTypes.string,
+    multi: PropTypes.bool,
     ...formInputProps,
     ...withMessagesProps,
   };
--- a/packages/input-select/src/Select.js
+++ b/packages/input-select/src/Select.js
@@ -112,8 +112,6 @@ class Select extends React.Component {
   normalizeItems = memoize(items => groupBy(items, "value"));
 
   renderTags = () => {
-    if (!this.props.multi) return null;
-
     const items = this.normalizeItems(this.props.items);
     return getValues(this.props)?.map(value => {
       const item = items[value]?.[0];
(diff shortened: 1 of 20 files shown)


packages/input-select/src/Select.js
```diff
@@ -28,7 +29,10 @@ class Select extends React.Component {
     super(props);
 
     this.state = {
-      selectedValue: props.value,
+      selectedValue:
+        props.multi && !Array.isArray(props.value)
```
alcferreira: I wonder if we should show a warning when it's `multi` and `value` is not an array because we could keep getting `value` as a primitive type instead of an array after component mounts 🤔

packages/input-select/src/Select.js
```diff
@@ -66,8 +70,13 @@ class Select extends React.Component {
   };
 
   onSelect = (selectedLabel, item) => {
+    const { onChange } = this.props;
     this.setState({ selectedLabel, isSearching: false });
-    this.props.onChange(item.value);
+    if (this.props.multi) {
+      onChange(xor(this.state.selectedValue, [item.value]));
```
alcferreira: Nice! 👌

packages/input-select/src/Select.test.js
```diff
@@ -204,6 +206,105 @@ describe("Select input", () => {
...
+
+      test.each`
+        currentValue | itemValue | expectedValue
+        ${[0, 1]}    | ${0}      | ${[1]}
+        ${[0, 1]}    | ${1}      | ${[0]}
+        ${[0, 1]}    | ${2}      | ${[0, 1, 2]}
+        ${[]}        | ${2}      | ${[2]}
+        ${[1, 2]}    | ${0}      | ${[1, 2, 0]}
+      `(
+        "should call onChange with $expectedValues when value is $currentValue and passed argument is $itemValue",
```
alcferreira:
```suggestion
        "should call onChange with $expectedValue when value is $currentValue and passed argument is $itemValue",
```

packages/input-select/src/Select.js
```diff
@@ -119,7 +123,20 @@ class Select extends React.Component {
...
+      >
+        {multi &&
+          items.map(
+            item =>
+              this.state.selectedValue.includes(item.value) && (
+                <Tag
+                  key={item.value}
+                  value={item.value}
+                  text={item.label}
+                  onRemove={this.onRemoveTag}
```
alcferreira: We are getting this error when an item is selected:
```
Warning: Unknown event handler property `onRemove`. It will be ignored.
    in div (created by Context.Consumer)
    in StyledComponent (created by styled.div)
    in styled.div (created by Tag)
```

The fix is pretty simple, would you mind adding the fix to this pr?
https://github.com/CraveFood/farmblocks/blob/master/packages/tag/src/components/Tag.js#L9-L11
Instead of passing props, we could pass only the remaining props to `StyledTag`

vNakamura: I did try that. The problem is that `StyledTag` needs the `onRemove` prop:
https://github.com/CraveFood/farmblocks/blob/6f28777edc7ce91370820da2493b58bcb925fd68/packages/tag/src/components/styledComponent/StyledTag.js#L6

alcferreira: Ouch lol

alcferreira: What about 

```jsx
<StyledTag rightPadding={onRemove? "0px" : "8px"} {...rest}>
```

vNakamura: Oh, the warning was just because it was an event handler name. Converting it to a boolean solves the problem.

packages/input-select/src/Select.js
```diff
@@ -194,23 +252,33 @@ class Select extends React.Component {
...
+    ...withMessagesProps,
     items: PropTypes.arrayOf(
       PropTypes.shape({
         value: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
         label: PropTypes.string,
         image: PropTypes.string,
       }),
     ),
-    value: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+    value: (props, ...rest) => {
```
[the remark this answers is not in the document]
vNakamura: That's how I did before https://github.com/CraveFood/farmblocks/pull/733#discussion_r267010806
This way it won't warn us if we use `multi` with a non-array `value`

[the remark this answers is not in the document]
MateusStanki: True, if we were using typescript as you would type this prop? hahaha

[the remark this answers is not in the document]
vNakamura: I did some research after we talked. It's actually doable in typescript:
```typescript
type Value = { multi: false, value: string | number }
type ValueArray = { multi: true, value: string[] | number[] }
type RegularProps = {
    items: {
        value: string | number,
        label: string,
        image: string,
    }
    . . .
}

type Props = RegularProps & (Value | ValueArray);

class Select extends React.Component<Props> { . . . }
```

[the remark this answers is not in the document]
MateusStanki: Nice, using this way, I think, we would have to use `Type Guards` to compiler infer types.

PR #800, opened 2019-05-21
Modal

### LINKED ISSUE #799: Create Modal component
Specs: https://share.goabstract.com/74d85b86-31fb-4da2-9eb5-b43a8f03ae72
image

vNakamura:
## Closes Issues

* Closes #799 

### Reviewer 1

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### Reviewer 2

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs


packages/modal/src/Modal.styled.js
```diff
@@ -0,0 +1,56 @@
...
+
+export const Wrapper = styled(animated.div)`
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  padding: 40px;
+  display: flex;
+  flex-direction: column;
+  justify-content: flex-start;
```
alcferreira: We can configure this setting by passing a `className` to the Modal and change it accordingly. 
Taking the `With custom parentNode` as an example, we would have this:

```diff
.add("With custom parentNode", () => {
    const Parent = styled.div`
      border: dashed 2px gray;
      width: 600px;
      height: 400px;
      padding: 10px;
      margin: 40px;
      position: relative;

+     .center {
+       justify-content: center;
+     }
    `;
    const Example = () => {
      const [isOpen, setIsOpen] = useState(false);
      const parentRef = useRef(null);

      return (
        <div>
          <button onClick={() => setIsOpen(true)}>Open Modal</button>
          <Modal
            parentNode={parentRef.current}
            isOpen={isOpen}
            onRequestClose={() => setIsOpen(false)}
+           className="center"
          >
            Hello.
          </Modal>

          <Parent ref={parentRef}>The Modal will render here</Parent>
        </div>
      );
    };
    return <Example />;
  });
```

Maybe we could allow this customisation via props, what do you think?

packages/modal/src/Modal.js
```diff
@@ -0,0 +1,126 @@
...
+            {slide.map(
+              ({ item: slideItem, key: slideKey, props: slideStyle }) =>
+                slideItem && (
+                  <animated.div key={slideKey} style={slideStyle}>
+                    <ConstrainedCard floating className="card" {...cardProps}>
+                      {showCloseButton && (
+                        <Header className="header">
+                          <Link
+                            className="close"
+                            rightIcon="wg-close-int"
```
alcferreira: It looks like something is cutting the icon... IDK if it's here on Modal or on Link

image

vNakamura: It's in the icons font

packages/modal/src/Modal.js
```diff
@@ -0,0 +1,126 @@
...
+  if (!parentNode) return null;
+
+  const handleKeyDown = event => {
+    if (shouldCloseOnEsc && event.key === "Escape") {
+      onRequestClose?.(event);
+    }
+  };
+
+  // eslint-disable-next-line consistent-return
+  useEffect(() => {
```
alcferreira: We could extract this to a custom hook

packages/modal/src/Modal.js
```diff
@@ -0,0 +1,126 @@
...
+  ContentWrapper,
+  Header,
+} from "./Modal.styled";
+
+const Modal = ({
+  isOpen,
+  parentNode,
+  shouldCloseOnOverlayClick,
+  shouldCloseOnEsc,
+  onRequestClose,
```
alcferreira: What about `onClose`?

packages/modal/src/Modal.js
```diff
@@ -0,0 +1,126 @@
...
+  children,
+  cardProps,
+  closeProps,
+  className,
+}) => {
+  if (!parentNode) return null;
+
+  const handleKeyDown = event => {
+    if (shouldCloseOnEsc && event.key === "Escape") {
+      onRequestClose?.(event);
```
alcferreira: Since the modal can be closed by clicking on the close icon, clicking on overlay and pressing ESC, I wonder if we should add a function that should be called when the modal gets closed `onModalClosed`

packages/modal/src/Modal.js
vNakamura: I think it would be misleading. Because the component is stateless, the events that trigger this prop don't actually close the modal. We have to manage the `isOpen` prop to make it close.

`onClose` could be called when the modal did close, after `isOpen` change to `false`. https://github.com/CraveFood/farmblocks/pull/800#discussion_r288758286

alcferreira: Hmm yeah, it makes sense. Actually we would never use this `onRequestClose`, right? Only if we didn't use the helpers to render the modal.

We are good here... Thanks for the clarification.

packages/modal/src/Modal.js
alcferreira: Right... So we should either update this icon on our fonticon or use another icon here

PR #835, opened 2019-07-11
Add option to customise the Tooltip position

### LINKED ISSUE #833: Add option to position Tooltip on top of its container
We need a Popover that opens on the top

alcferreira:
## Closes Issues

* Closes #833 

### Reviewer 1

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### Reviewer 2

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs


packages/tooltip/src/Tooltip.js
```diff
@@ -61,19 +62,21 @@ Tooltip.propTypes = {
   children: PropTypes.node,
   isVisible: PropTypes.bool,
   align: PropTypes.oneOf(values(alignments)),
+  position: PropTypes.oneOf(values(position)),
```
vNakamura: Since we're already having a breaking change for `top`->`offset`, how about renaming the `align` and `position` props?
They're the same thing but for different axis. The naming doesn't make it clear which one is horizontal and which one is vertical.
We could go with something like `horizontalAlign`/`verticalAlign` or shorter, `hAlign`/`vAlign`. What do you think?

alcferreira: It's better indeed, thank you! 
I would go with `hAlign`/`vAlign`, but what about `alignX`/`alignY`?

alcferreira: I'll wait for the complete review before I do this change, ok?

vNakamura: I like alignX/Y!

packages/tooltip/src/Tooltip.story.js
```diff
@@ -1,137 +1,249 @@
...
+        {children}
+      </div>
+    );
+    return (
+      <div>
+        <BoundaryContainer className="exampleContainer">
+          <Tooltip
+            trigger={<Button type={buttonTypes.SECONDARY}>Hover me</Button>}
+            content="Auto aligned/positioned tooltip"
+            positionX={POSITIONS.X.AUTO}
```
MateusStanki: I think that we can simplify the position API, something like `placement`: `top`. 'topRight', 'bottomLeft'. What do you think?

packages/tooltip/src/Tooltip.story.js
```diff
@@ -1,137 +1,249 @@
...
+      margin: "40px",
+      border: "none",
+    };
+    return (
+      <div style={{ display: "flex", flexDirection: "column", marginTop: 60 }}>
+        <div style={{ display: "flex", justifyContent: "space-between" }}>
+          <div style={marginStyle}>
+            <Tooltip
+              trigger={<Button type={buttonTypes.SECONDARY}>Hover me</Button>}
+              content="Bottom left aligned"
```
MateusStanki: Change `content` 'bottom' word on **top positioned** tooltip stories

packages/tooltip/src/Tooltip.story.js
alcferreira: I see that some libs are going that way, but for our use case I personally prefer the way it was implemented because we can set the position on each axis without considering the position of the other one.
We would have to make some tweaks to work the way you proposed.

What's your opinion @vNakamura?

vNakamura: I'm good with two props.
In most cases, we would just set `positionX`. We never needed the tooltip at the top until now.

MateusStanki: Ok, I prefer an enum of string that covers all the cases, easier to use and read. But if this way is ok for you guys, I'm ok too!

packages/tooltip/src/Tooltip.js
```diff
@@ -1,79 +1,97 @@
...
-      (bondariesNode && bondariesNode.getBoundingClientRect().right) ||
-      window.innerWidth;
+  .appear-enter {
+    opacity: 0;
+  }
 
-    if (right >= maxRight) return alignments.RIGHT;
+  .appear-enter-active {
+    opacity: 1;
+    transition: opacity 100ms ease-in;
```
vNakamura: Same from Popover. I think we should use at least 200ms.

packages/tooltip/src/Tooltip.story.js
```diff
@@ -1,137 +1,249 @@
...
-      <Tooltip
-        css="
-          text-decoration-line: underline overline;
-        "
-        isVisible
-        text="This is a left aligned tooltip"
-      />
-    </div>
-  ));
+  .add("auto aligned/positioned (with boundaries)", () => {
```
vNakamura: The auto position is not working here.

PR #859, opened 2019-07-30
Tooltip auto positioned on Y axis should consider the tooltip height

### LINKED ISSUE #858: Tooltip auto positioned on Y axis should consider the tooltip height
The tooltip autopositioned on Y axis only renders on top if the start of the tooltip is out of boundaries. Actually it would be better to checkt if the bottom of the tooltip is out of boundaries.

Example:

alcferreira:
## Closes Issues

* Closes #858

### Reviewer 1

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### Reviewer 2

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### DIFF
--- a/packages/tooltip/src/Tooltip.story.js
+++ b/packages/tooltip/src/Tooltip.story.js
@@ -131,6 +131,7 @@ storiesOf("Tooltip/Tooltip", module)
               content="Top center aligned without arrow"
               hideArrow
               positionX={POSITIONS.X.CENTER}
+              positionY={POSITIONS.Y.TOP}
             />
           </div>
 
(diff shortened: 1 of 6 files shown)


packages/tooltip/src/Tooltip.story.js
```diff
@@ -176,7 +177,7 @@ storiesOf("Tooltip/Tooltip", module)
           justifyContent: "space-between",
           width: "100%",
           position: "fixed",
-          bottom: 0,
+          bottom: 30,
```
vNakamura: The bottom right button in this story is slightly out of bounds:
image

I think it's caused by the `position: fixed`
We can remove it and use this on that `div` at line 152:
```jsx
<div
  style={{
      display: "flex",
      flexDirection: "column",
      justifyContent: "space-between",
      height: "calc(100vh - 30px)",
  }} 
>
```

alcferreira: fixed, thanks

PR #910, opened 2019-09-19
[Amount selectors] Add tooltip when user types a value above the maximum amount allowed

### LINKED ISSUE #907: [Amount selectors] Render a tooltip warning about limits
- [x] We need to add a tooltip below the input that should be visible whenever user types in an amount outside limits.
- [x] Fix bug where the component accepts a value outside limits

- [ ] ~Limit initial value if the availability is less than 1.~ Actually we need the ability to keep an initial value outside boundaries. We just adjust automatically after an user action

alcferreira:
## Closes Issues

* Closes #907

### Reviewer 1

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### Reviewer 2

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### DIFF
--- a/packages/amount-selectors/src/components/AmountSelectors.js
+++ b/packages/amount-selectors/src/components/AmountSelectors.js
@@ -1,9 +1,11 @@
 import * as React from "react";
 import PropTypes from "prop-types";
+import values from "object.values";
+import { Transition } from "react-spring/renderprops.cjs";
 import Button, { buttonTypes, buttonSizes } from "@crave/farmblocks-button";
 import InputText from "@crave/farmblocks-input-text";
 import { fontSizes } from "@crave/farmblocks-theme";
-import values from "object.values";
+import { TooltipContent, POSITIONS } from "@crave/farmblocks-tooltip";
 
 import selectorSizes from "../constants/selectorSizes";
 import Wrapper from "../styledComponents/AmountSelector";
@@ -31,6 +33,7 @@ class AmountSelectors extends React.Component {
       disableBoth: false,
       tooltipText: "",
       displayValue,
+      focused: false,
     };
   }
 
@@ -47,9 +50,7 @@ class AmountSelectors extends React.Component {
 
   onChange = (event, cb) => {
     const value = typeof event === "number" ? event : event.target.value;
-    const hasBrowserValidation = !!(
-      event.target && event.target.validity !== undefined
-    );
+    const hasBrowserValidation = event?.target?.validity !== undefined;
     const disableBoth =
       hasBrowserValidation && this.disableBoth(event.target.validity);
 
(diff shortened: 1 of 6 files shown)


packages/amount-selectors/src/components/AmountSelectors.js
```diff
@@ -156,19 +192,22 @@ class AmountSelectors extends React.Component {
     step: PropTypes.number,
     min: PropTypes.number,
     max: PropTypes.number,
+    maxAmountMessage: PropTypes.string,
```
vNakamura: Shouldn't we have a message for the minimum amount as well?

alcferreira: I don't think we have a use case today where a min amount message would be useful, so I decided to implement the max amount only

vNakamura: I don't get it. The use case is the same for the maximum: The user types a value that is out of the acceptable range.
On the webapp, a pack size with whole increments would have a minimum of 1. If the user types `0,5`, it would not show any message but it will change back to 1 when the user changes focus. They may not notice this change and send the order with an unexpected amount.

alcferreira: Yep, you're right.. I missed this point, thanks! I'll submit some changes to handle this

alcferreira: done

packages/amount-selectors/src/components/AmountSelectors.js
```diff
@@ -102,7 +111,24 @@ class AmountSelectors extends React.Component {
...
+      max,
+      min,
+      minAmountMessage,
+      maxAmountMessage,
+      showBoundariesMessageOnlyOnFocus,
+    } = this.props;
+    const { focused, value } = this.state;
+    const showMessage =
+      !showBoundariesMessageOnlyOnFocus ||
+      (showBoundariesMessageOnlyOnFocus && focused);
```
MateusStanki: You don't need to check if `showBoundariesMessageOnlyOnFocus` is Truly here, you already checked it in the previous line.

packages/amount-selectors/src/components/AmountSelectors.js
```diff
@@ -102,7 +111,24 @@ class AmountSelectors extends React.Component {
...
+    const showMaxMessage = value > max && showMessage;
+    const showMinMessage = value < min && showMessage;
```
MateusStanki: When the user deletes the input number, we are showing the `minAmountMessage`, check if the input is empty before show the message.

MateusStanki: If I enter a number then a point, the buttons become disabled and show `minAmountMessage `
Screen Shot 2019-09-27 at 11 23 43

packages/amount-selectors/src/components/AmountSelectors.js
```diff
@@ -47,9 +50,7 @@ class AmountSelectors extends React.Component {
 
   onChange = (event, cb) => {
     const value = typeof event === "number" ? event : event.target.value;
-    const hasBrowserValidation = !!(
-      event.target && event.target.validity !== undefined
-    );
+    const hasBrowserValidation = !!(event?.target?.validity !== undefined);
```
MateusStanki: don't need `!!` as it is a comparison operator

alcferreira:
@MateusStanki changes addressed, tks

PR #914, opened 2019-09-26
Selected image class

### LINKED ISSUE #913: [Select]:  Differentiate selected item image class
Differentiate the selected item image class from the dropdown items image class.

MateusStanki:
## Closes Issues

* Closes #913

### Reviewer 1

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### Reviewer 2

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### DIFF
--- a/packages/input-select/src/components/withImage.js
+++ b/packages/input-select/src/components/withImage.js
@@ -20,7 +20,7 @@ export default Input =>
         <LabelContainer className="labelContainer">
           {image && (
             <ItemImage
-              className="itemImage"
+              className="selectedItemImage"
               src={image}
               size={badgeSizes.SMALL}
               badge
--- a/packages/input-select/src/Select.story.js
+++ b/packages/input-select/src/Select.story.js
@@ -18,7 +18,7 @@ const moreItems = [
   { value: "6", label: "Coconut" },
 ];
 
-const imgSrc = "https://picsum.photos/100?image=0";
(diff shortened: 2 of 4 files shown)


packages/input-select/src/components/withImage.js
```diff
@@ -20,7 +20,7 @@ export default Input =>
         <LabelContainer className="labelContainer">
           {image && (
             <ItemImage
-              className="itemImage"
+              className="selectedItemImage"
```
alcferreira: just changed the classname here... Is it enough?

MateusStanki: That's the pr purpose.

alcferreira: Still didn't get based on the issue description. Can you elaborate more?

MateusStanki: Yes, for sure. In select inputs we have the dropdown items which are the options, and the item that was selected (if there is one). We have few different styles in both, mainly using the small size. The problem is that in both cases we use the same class, which is not used internally, was created to have access to the element style at implementation time. Applying styles in this class gave different behaviors between the selected item and the others. So, doesn't make any sense to use the same class on both, correct? That's why i created this simple PR.

alcferreira: Ohhhh now i see, i thought it was about the selected image inside the dropdown items
Thanks!

PR #920, opened 2019-10-02
Create PhoneInput component

### LINKED ISSUE #912: New component - Phone Input
## Requirements
- the value should be in RFC3966 (e.g. `tel:+12133734253`) but it should be presented to the user accordingly to the country's standard (e.g. in the US: `(213) 373-4253`).
- fallback to the [2 letter code](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) when emoji is not supported by the browser.
- accept a list of countries that are most likely to be selected. Those should be presented at the top, but they should also be available on their proper position in alphabetical order.
- accept a prop for the default country. Changes in this prop should reflect in the component as long as the phone number is empty

## Resources
- https://github.com/catamphetamine/libphonenumber-js
- https://www.npmjs.com/package/detect-emoji-support
- https://github.com/mledoze/countries/blob/master/countries.json
- https://fusejs.io/
- https://github.com/bvaughn/react-window

## Design specs:
https://share.goabstract.com/24942666-c3fc-421c-a373-42d9f8860af1
image
image

vNakamura:
## Closes Issues

* Closes #912 

### Reviewer 1

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### Reviewer 2

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs


packages/input-phone/src/PhoneInput.js
```diff
@@ -0,0 +1,309 @@
...
+        />
+      }
+      active={popoverOpen}
+      value={formatIncompletePhoneNumber(phone?.format("NATIONAL"), country)}
+      onChange={handleNumberChange}
+      inputMode="tel"
+      css={`
+        .input {
+          overflow: unset;
+          @media only screen and (max-width: ${fullScreenBreakpoint}) {
```
alcferreira: I believe here should be `tooltipProps?.fullScreenBreakpoint || fullScreenBreakpoint`

As `fullScreenBreakpoint` is primarily a `tooltip` prop and we are passing down `tooltipProps` to the tooltip component, it makes sense to check it here

packages/input-phone/src/PhoneInput.hooks.js
```diff
@@ -0,0 +1,108 @@
...
+const options = {
+  shouldSort: true,
+  keys: ["name", "nativeNames", "altSpellings", "flag", "callingCode"],
+};
+
+export const useCountrySearch = (data, query, priority) => {
+  const fuse = useRef(new Fuse(data, options));
+
+  const allCountries = useMemo(() => {
+    const codes = priority?.split(",");
```
alcferreira: I guess we should make sure this piece of code will never break either by wrapping it with a try catch block or by checking if `priority` is a string

Nothing presents user from passing an array, by mistake (or not lol), then it would break..

packages/input-phone/src/PhoneInput.js
```diff
@@ -0,0 +1,310 @@
...
+            return (
+              // eslint-disable-next-line jsx-a11y/no-static-element-interactions, jsx-a11y/click-events-have-key-events
+              <div
+                onClick={event => {
+                  // We stop propagation to avoid giving focus to the main input
+                  // This happens because the popover is inside the input wrapper
+                  event.stopPropagation();
+                }}
+                css="
+                display: flex;
```
MateusStanki: I think it will be good to extract this and the `ul` bellow to styled-components.
PhoneInput is already heavily loaded, to improve this maybe you can decompose all the `content` render to another component or create a container to handle all these logic, which will make a good lean.

packages/input-phone/src/PhoneInput.js
```diff
@@ -0,0 +1,310 @@
...
+                    )}
+                  </AutoSizer>
+                </ul>
+              </div>
+            );
+          }}
+        />
+      }
+      active={popoverOpen}
+      value={formatIncompletePhoneNumber(phone?.format("NATIONAL"), country)}
```
MateusStanki: extract to a const and memoize to change only when the phone or country changes

packages/input-phone/src/CountrySelectorTrigger.js
```diff
@@ -0,0 +1,32 @@
...
+
+const Trigger = styled.div`
+  padding: 0 4px 0 8px;
+  height: 100%;
+  display: flex;
+  align-items: center;
+  cursor: ${({ disabled }) => (disabled ? "default" : "pointer")};
+`;
+
+const CountrySelectorTrigger = ({ callingCode, flag, code, ...props }) => {
```
MateusStanki: Use React.memo here the component is fully stateless

packages/input-phone/src/CountrySelectorTrigger.js
```diff
@@ -0,0 +1,32 @@
...
+  return (
+    <Trigger {...props}>
+      {emojiSupport() ? flag : `${code} `}+{callingCode}
```
MateusStanki: `emojiSupport` can be called on mount only

vNakamura: `detect-emoji-support` already caches the results:
https://github.com/danalloway/detect-emoji-support/blob/0ceca07e9fd51077f372fd15d99eb0e4a142083b/lib/index.js#L3-L6

packages/input-phone/src/PhoneInput.hooks.js
```diff
@@ -0,0 +1,118 @@
...
+          return;
+        }
+
+        selectFn(items[highlightIndex].code);
+        return;
+      }
+
+      if (key === "Escape") {
+        setHighlightIndex(-1);
+        cancelFn?.();
```
MateusStanki: `cancelFn` isn't available on open. We need to query something or browse the list to close the popover on "esc"

packages/input-phone/src/PhoneInput.js
vNakamura: I tried extracting the whole popover content, but it needed so many props that it made it worse. I ended up extracting just the `ul`.

packages/input-phone/src/PhoneInput.hooks.js
vNakamura: I closured the `.current` instead of the ref

vNakamura:
@MateusStanki changes addressed. Thanks!

PR #1036, opened 2021-03-29
fix(Checkbox): prevent checkbox from shrinking 

### LINKED ISSUE #1033: Checkbox should not shrink when there's not enough width
image

juliansotuyo:
## Closes Issues

* Closes #1033

### Reviewer 1

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### Reviewer 2

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs


packages/input-checkbox/src/styledComponents/Checkbox.js
```diff
@@ -23,8 +23,8 @@ const checkboxColors = props => {
 const checkboxStyles = props =>
   !props.switch &&
   css`
-    width: ${checkboxSize};
-    height: ${checkboxSize};
+    max-width: ${checkboxSize};
+    max-height: ${checkboxSize};
```
alcferreira, line 27: I believe the correct prop here would be to set the min-width and min-height, right?

See how it still shrinks with the max props:
Screen Shot 2021-04-09 at 17 25 39

alcferreira (changes requested):

packages/input-checkbox/src/styledComponents/Checkbox.js
juliansotuyo: So, I replaced the max with min and at first it worked, but when I opened the webapp on the phone the checkbox was huge. To fix it I added the props `min-width` and `min-height` together with`max-width` and`max-height`

alcferreira (approved):

PR #1045, opened 2021-05-05
Responsive carousel component

### LINKED ISSUE #1043: Carousel component

juliansotuyo:
## Closes Issues

* Closes #1043 

### Reviewer 1

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs

### Reviewer 2

* [ ] Functions properly
* [ ] Has tests
* [ ] Has docs


alcferreira (changes requested):
Cool! Here are some points that I found out while testing it... Worth mentioning that I haven't reviewed the code just yet

1. we should add a default value for qtyOfSlidesPerSet so it won't break the app when it's undefined
image
2. `qtyOfSlidesPerSet` won't work with a number, it only works with an array. When we pass a number, it always renders 2 slides per set
image
3. `qtyOfSlidesPerSet` doesn't seem to be working with an array set as well.. It seems like it always get the second item from the array and use it disregarding the screen size
4. We should be able to customize some slide props like border-radius
5. The arrow buttons should be `small` by default, right?
6. Should we be able pass along left/right button on our own? Say we would like to pass a different icon or different button, should we be able to?
7. Something is off with the `infiniteLoop`

packages/carousel/src/Carousel.js
```diff
@@ -21,6 +21,8 @@ function Carousel({
   breakpoints,
   children,
   style,
+  leftButton,
+  rightButton,
```
alcferreira, line 25: Let's remove ambiguity with `leftButtonProps` and `rightButtonProps`

packages/carousel/src/Carousel.js
```diff
@@ -145,7 +147,8 @@ function Carousel({
           {showLeftArrow && (
             <ArrowButton
               data-testid="left-arrow"
-              icon={<SmChevronLeft size={24} />}
+              {...leftButton}
+              icon={leftButton?.icon || <SmChevronLeft size={24} />}
```
alcferreira, line 151: if we pass props after the default props, we don't have to to this condition `icon={leftButton?.icon || <SmChevronLeft size={24} />}`

packages/carousel/src/Carousel.js
```diff
@@ -194,8 +198,10 @@ Carousel.propTypes = {
       slidesToShow: PropTypes.number,
     }),
   ),
-  style: PropTypes.oneOfType([PropTypes.object, PropTypes.array]),
   infiniteLoop: PropTypes.bool,
+  style: PropTypes.oneOfType([PropTypes.object, PropTypes.array]),
```
alcferreira, line 202: why do we need a prop called style? 🤔

alcferreira:
We will work on another version of this component soon

